PR #69, opened 2023-11-06
[JA] translate new files

tequdev:


public/locales/ja/common.json
```diff
@@ -63,22 +63,22 @@
...
+      "02": "私のテーブル"
     }
   },
   "signin": {
     "signin": "サインイン",
     "signout": "サインアウト",
     "choose-app": "サインインするアプリを選択",
     "login-with": "Login with",
-    "sign-with": "Sign with",
+    "sign-with": "Login with",
```
ihomp, line 75: Login - is just to get an address and show information 
Sign - is to sign the transaction with Xumm or hardware wallet etc. To use a device/wallet to sign it. (to confirm)

tequdev: This is just my mistake😢

PR #84, opened 2024-01-08
Add basic page for transaction

pavel-fokin:


ihomp (approved):

ihomp:
we can merge it, it's a new folder, a new page.. it does not conflict with anything. 
I just can't merge drafts :) 

The only thing I found, is the **explorerLink**, with that transaction page live we should get away from the old Explorer pages 
so if it's transaction link it will /transaction/hash if it's an address it will be /account/address, if it is nft it will be /nft/nftid

ihomp:
Awesome job, by the way, you are a better coder than me :) neat code

pavel-fokin:
Sure, we can merge 🙏 I made some fixes and change link to the account instead of old explorer.

PR #93, opened 2024-01-16
Add transaction view for payment type

pavel-fokin:


utils/transaction.js
```diff
@@ -1,7 +1,7 @@
 import moment from "moment";
 
 const formatDateTime = (date) => {
-  return moment(date).utc().format("YYYY-MM-DD HH:mm:ss");
+  return moment(date).format("YYYY-MM-DD HH:mm:ss");
```
ihomp, line 4: we have a function to show date and time
`import {fullDateAndTime } from '../utils/format'`
it shows the date and time in locally for user format

components/Transaction/Links.js
```diff
@@ -1,22 +1,25 @@
-import Link from 'next/link'
+import Link from "next/link";
 
 const LinkExplorer = ({ hash }) => (
-    <a href={`https://bithomp.com/explorer/${hash}`} target="_blank" rel="noopener noreferrer">
-        {hash}
-    </a>
-)
+  <a
+    href={`https://bithomp.com/explorer/${hash}`}
```
ihomp, line 5: we need to replace old Explorer links to 
new links like 
/tx/
/nft/
/account/
etc

components/Transaction/TransactionDetails.js
```diff
@@ -61,54 +39,61 @@ export const TransactionDetails = ({ tx }) => {
...
+              <Table.Data>
+                <LinkAccount address={tx.address} />
+              </Table.Data>
+            </Table.Row>
+            <Table.Row>
+              <Table.Data>Sequence:</Table.Data>
+              <Table.Data>#{tx.sequence}</Table.Data>
+            </Table.Row>
+            <Table.Row>
+              <Table.Data>XRLP fee:</Table.Data>
```
ihomp, line 85: XRPL :)

components/Transaction/TransactionDetails.js
```diff
@@ -61,54 +39,61 @@ export const TransactionDetails = ({ tx }) => {
...
+              <Table.Data>{tx.outcome?.fee}</Table.Data>
```
ihomp, line 86:
```
import {amountFormat } from '../utils/format'

{amountFormat(tx.outcome?.fee)}
```

components/Transaction/TransactionPayment.js
```diff
@@ -1,25 +1,26 @@
...
 
-    const source = tx.specification?.source?.address
-    const destination = tx.specification?.destination?.address
+export const TransactionPayment = ({ tx }) => {
+  const destination = tx.specification?.destination?.address;
 
-    return (
-        <TransactionDetails tx={tx}>
-            <tr>
-                <td>Source:</td>
```
ihomp, line 12: Source was removed ? :)

utils/transaction.js
pavel-fokin: `fullDateAndTime` returns `Invalid Date` for `'2023-12-13T13:42:21.000Z'`. I'll localize timestamps a bit later. Let's keep as is for now.

components/Transaction/TransactionDetails.js
pavel-fokin: `amountFormat` gives `1.2e-11 XRP` for low values, it looks a bit not user friendly.

pavel-fokin:
I fixed most of the comments. But I'll close this PR as it is included to the https://github.com/Bithomp/frontend-react/pull/94.

utils/transaction.js
ihomp: it expects the Unix timestamp I guess, so we rather update what endpoint returns from the backend, as we usually return timestamps in /v2/ 

Or as an option we can update the fullDateAndTime to accept the '2023-12-13T13:42:21.000Z' format too. 

You can leave it for now, but would be good to leave a comment so I won't forget to update later :)

PR #125, opened 2024-04-16
add: filters in nft-explorer

Vierin:


components/AddressInput.js
```diff
@@ -0,0 +1,143 @@
...
+
+import {
+  isValidCTID,
+} from '../utils'
+
+import { userOrServiceLink } from '../utils/format'
+
+let typingTimer
+
+export default function AddressInput({ searchPlaceholderText, userData = {}, setFilters, type, inputValue, title, link }) {
```
ihomp, line 14: userData - this is usually an authorized user's address. 

addressInput is intended to be used for fields like 
receiver, issuer, and seat—so most likely it won't be the same user as the one that is logged in. 
Therefore, probably won't need here userData.

components/AddressInput.js
```diff
@@ -0,0 +1,143 @@
...
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
+
+  useEffect(() => {
+    if (userData?.address) {
+      setSearchItem(userData.address)
+    }
+  }, [userData])
```
ihomp, line 33: so probably we won't need this too

components/AddressInput.js
```diff
@@ -0,0 +1,143 @@
...
+      return
+    }
+
+    //if more than 3 characters - search for suggestions
+    let notEmpty = false;
+    if (value && value.length > 0) {
+      clearTimeout(typingTimer)
+      typingTimer = setTimeout(async () => {
+
+        if(type === 'search' || type === 'issuer') {
```
ihomp, line 49: What are the differences between a search and an issuer?

public/locales/en/common.json
```diff
@@ -363,6 +363,7 @@
   },
   "table": {
     "mints": "Mints",
+    "view": "View",
```
ihomp, line 366: if we add something to the common.json or any already existing language files, that should be added to all language files. 
adding an English version to all languages is fine, except Russian

components/AddressInput.js
Vierin: 'search' is a type of name input, I changed to the 'name' to make it more clear

ihomp: then issues isn't very clear either:) 

user can type in
a username 
an address
a xaman slug
a globaild slug
a service name 

if an address - then it can be `address` - if not an address (all the rest) it can be `name`

Vierin: I removed this part of the code to make it clearer

PR #135, opened 2024-05-31
Filter: add new view toggle

Vierin:


ihomp (changes requested):
There are lots of conflicts to solve, it's better to move the specific changes to a new branch for an easier review.

PR #146, opened 2024-07-07
All issuers

Vierin:
Changed place for All issuers select 
before: 
image 
after: 
image

### DIFF
--- a/components/NftsComponent.js
+++ b/components/NftsComponent.js
@@ -720,7 +720,7 @@ export default function NftsComponent({
               </div>
             }
 
-            {!nftExplorer && (id || owner) && issuersList?.length > 0 && rendered &&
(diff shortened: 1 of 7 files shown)


ihomp:
Screenshot 2024-07-08 at 17 05 53
Screenshot 2024-07-08 at 17 05 59

in the dark mode..

ihomp:
It would be better not to hide the issuer field, but show it disabled.
I've added the functionality, but styles seems off.. you can't see it is disbaled. would be nice to show it with disbaled styles, like other elements have. 

Taxon looks disabled, CSV button looks disabled, but not the dropdown

Screenshot 2024-07-08 at 17 15 50
Screenshot 2024-07-08 at 17 16 38

ihomp:
Super nice !

PR #162, opened 2024-08-21
fix: nft scroll then filter open on mobile

Vierin:


ihomp:
I've tested on the chrome mobile version - no changes.

PR #163, opened 2024-08-25
fix: filters scrollbar on nft-volumes page

Vierin:
removed the scrollbar in filters and added a border to separate sections
before:
image
after:
image


ihomp:
- it has hydration errors as it is based on the old branch. 
- Where do you see the scroll? windows? which browser? 
- I do not see scroll on chrome, mac os

PR #165, opened 2024-08-25
fix: chart resizing on filters toggle

Vierin:
before: 
image
after:
image


ihomp:
PR is deleting lots of the recent staff. 
PR needs to be sent from the latest branch.

PR #169, opened 2024-09-01
Products section on home page

Vierin:


ihomp (commented):
1. there are changes from another commit here
2. would be nice on bigger screen to have all 4 in one line 
3. highlighting of the block can be confusing as each item inside each block will be a link. 
4. most likely we will update the pics, would be easier to update them if the don't have the green/blue light on them (as on the first one with nfts) the light spot is ok to have, but it would be better to add it to the background as base64 svg or css.

PR #175, opened 2024-09-09
faucet translations

Anna15170221:


ihomp:
need to check that all are correct, I see that 

`Confirmed within <2>{{time}} seconds</2>`
become
`Confirmed within <`

That will brake translation.

PR #182, opened 2024-09-18
add: products + fix gradient bg

Vierin:


ihomp:
Nice! 

would it be possible to have 4 in a row for the most common mac resolutions 14'' 15'' 

I think I have 1523 px

so, for 1520px can we make 4 in one row?

ihomp:
the first one doesn't have the higlight animation. 
may be we can just remove it on others too, so it will be the same?

ihomp:
Instead of making the whole cart as a link, we would need to make each listed point as a link.

PR #189, opened 2024-11-14
preview images

Anna15170221:
preview images
300x157 is also added for X


ihomp (changes requested):
The images need to be renamed, the name should be exactly the same as the page slug in the browser when you open that page. 
the correct pages names can be also looked here: 
https://github.com/Bithomp/frontend-react/tree/main/pages

@Anna15170221

PR #193, opened 2024-11-19
Xahauexplorer icon

Anna15170221:
square version for better look


ihomp:
The size increase is to large +1200% 
+163 KB (1200%) 

I'll cancel this PR, and convert icon myself.

PR #210, opened 2025-02-06
fixed, deleted and added tokens

go140point6:
# Token List Update

Forgive me if this isn't in the format expected; I've never really collaborated via a pull request before.

I audited my current token list and found several errors in Koinly IDs, as noted below. I also removed some of the XAHL tokens that were duplicated on the XRPL side and added the RLUSD token.

## Updated Tokens - XRPL

- **$BAE**
- **CLV**  
  - Based on `issuer.currency`, the XRPL token Koinly ID was updated.
- **DVT**
- **EVR (Gatehub)**  
  - Previously, I was using the native EVR (XAHL side) Koinly ID. This update now uses the Koinly ID assigned to the Gatehub EVR.  
  - **Note:** Koinly doesn't appear to recognize the Gatehub XAH token. There seem to be hundreds of fake XAH tokens assigned Koinly IDs, making it impossible to find the real Gatehub XAH token. The `issuer.currency` "trick" doesn't work.
- **LP Token CSC-XRP**  
  - Previously used the generic "LP Token 3."
- **LP Token EQ-XRP**  
  - Previously used the generic "LP Token 4."
- **LP Token XAH-XRP**  
  - Previously used the generic "LP Token 5."
- **Note:**  
  - Koinly doesn't appear to recognize "LP Token EVR-XRP," so I left that as the generic "LP Token 6."

## Deleted Tokens - XRPL

- **EVR (native)**  
  - This was inadvertently duplicated on the XRPL side; it belongs only on the XAHL side.
- **PXP**  
  - This was inadvertently duplicated on the XRPL side; it belongs only on the XAHL side.

## Added Token - XRPL

- **RLUSD**

---

### A Note About the XAHL Side

The two minor tokens I know about, **CLV** and **PXP**, have no corresponding Koinly IDs. As a placeholder, I simply picked tokens with the same symbol. I'm not sure what the "correct" approach would be in this case but it is very common for a token not to have a Koinly ID and then suddenly to have one.


ihomp:
Thank you very much!

PR #228, opened 2025-03-18
combine sales and volumes chart

juddydev:
I've combined sales and volumes chart into 1 chart

image


ihomp:
It would be better to reuse the component, the issue is already resolved in a different PR.

PR #229, opened 2025-03-19
Blacklisted accounts article

Anna15170221:


ihomp (changes requested):
- Cannot test; there is an error: `Expected server HTML to contain a matching <li> in <p>.`
- remove unnessary spaces 
- rename page and main finction to to blacklisted-address and Blacklistedaddress
- probably images are not the same width (not tested) but they should be the same width 
- XRP should be replaced with nativeCurrency, XRPL with ledgerName from util.js

PR #234, opened 2025-03-25
Fix searchbar

Harastar1001:
- I fixed error with searchbar


ihomp:
@Harastar1001 Same thing, this PR has commits from other issues. 
We need PR to have code changes only for one issue. 
I've answered how to do it in another PR.

Harastar1001:
Okay, got it.
I will fix it now.

PR #239, opened 2025-03-26
fix: new pr for searchbar

Harastar1001:
Hello, I fixed Searchbar

### DIFF
--- a/components/Layout/SearchBlock.js
+++ b/components/Layout/SearchBlock.js
@@ -21,7 +21,7 @@ import {
 import { userOrServiceName, amountFormat } from '../../utils/format'
 
 import { IoSearch } from 'react-icons/io5'
-
+import { IoMdClose } from 'react-icons/io'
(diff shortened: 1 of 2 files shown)


components/Layout/SearchBlock.js
```diff
@@ -323,6 +323,12 @@ export default function SearchBlock({ searchPlaceholderText, tab = null, userDat
     return ''
   }
 
+  const clearAll = () => {
+    if (setSearchItem) {
```
ihomp, line 327: Are you checking if the function is defined here, or did you want to check that there is a value `searchItem` ?

ihomp:
It's better to keep t function where it was before, as we have it that way in all other files; otherwise, it needs to be replaced everywhere in the project.

ihomp:
On hover on dark mode the color is changing

ihomp (changes requested):

PR #240, opened 2025-03-26
fix error message when it's not find nfts

Harastar1001:
This is a PR related to the error message that appears when nft is not found.


ihomp (changes requested):
It's better to keep t function where it was before, as we have it that way in all other files; otherwise, it needs to be replaced everywhere in the project.

PR #243, opened 2025-03-27
Show nft-mint page for xrpl network

juddydev:


ihomp:
The code is only changing the transaction type. 
We need all the new fields described in the ticket.

ihomp (changes requested):
- We need two separate components one for xrpl - nftokenmint another one for xahau - uritokenmint
- we do not need the network tabs on that page 
- there is no digest on xrpl.
- we need different foelds on xrpl as in the ticket.

juddydev:
Could  you share more details ?
I need to create input element that user can fill require field ?

ihomp:
> Could you share more details ? I need to create input element that user can fill require field ?
> 
> It mean that I need to create
> 
> * Input for user input TransferFee
> * Input for user fill Issuer
> * Input for user fill URI
> 
> ...
> 
> * Select for User choose Flags

As described in the ticket, we have similar fields in other pages, the components can be re-used.

ihomp:
1) The structure of `nft-mint.js` should be very simple; it should load a component depending on the network type.
```
        {xahauNetwork ? (
          <UritokenMint setSignRequest={setSignRequest} />
        ) : (
          <NftokenMint setSignRequest={setSignRequest} />
        )}
```

right now, you have one component on the page right away; another one is loading 

2) the components shouldn't be in the `pages` 
right now you've added a component to `/pages/services/xrpl-nft-mint.js`
Instead, you need to have two new components:
`/components/Services/NftMint/UritokenMint.js`
`/components/Services/NftMint/NftokenMint.js`

where Uritorkenmint is the NFT mint for xahau, and NftokenMint is the NFT mint for xrpl.

ihomp:
Regarding the page:
- The transfer fee should be in percent; it's not user-friendly otherwise. 
- The issuer field could be an address-search field instead, as on other pages. 
- it's better to pass 'account' from parents rather than take from localStorage, as in other components/pages 
- Transfer fee maxLength="256" ? thats too much according to the specs 
- Taxon - same thing, 256 is too much
- we have a taxon validation `isValidTaxon`, in `utils.js ` it's need to be used to validate taxons 
- No amount conversion; it's not XRP, it's drops! 
- Destination needs to be as address-search fields as described in the ticket 
- Flags - shouldn't be a single-select; user should be able to choose all 4 flags. 
- Expiration should be optional, with an option "no expiration". 

There is no need to continue working on the ticket. Thank you for your time @juddydev

ihomp (commented):
It is far away from being approved; the PR can be dropped.

PR #245, opened 2025-03-28
export csv for coin ledger

juddydev:


pages/admin/pro/history.js
```diff
@@ -440,32 +498,34 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
...
                 >
-                  <DownloadIcon /> CSV for Koinly
+                  <DownloadIcon /> CSV for {platformCSVExport}
                 </CSVLink>
               )}
+              {/* <br />
               <br />
-              <br />
-              Let us know if we miss koinlyIDs for your tokens. We will add them to the system.
+              Let us know if we miss koinlyIDs for your tokens. We will add them to the system. */}
```
ihomp, line 528: Why did you comment it? the text should be visible when koinly is chosen

pages/admin/pro/history.js
```diff
@@ -252,10 +310,10 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
 
         res.activities[i].timestampExport = new Date(res.activities[i].timestamp * 1000).toISOString()
 
-        res.activities[i].sentAmount = sending ? res.activities[i].amountNumber : 0
+        res.activities[i].sentAmount = sending ? res.activities[i].amountNumber : ""
```
ihomp, line 313: that is a dangerous chnage that can brake csv for koinly.. was that tested?

juddydev: I have tested this
We can import it into Koinly
I have imported into Koinly successfully !
image


[koinly.csv](https://github.com/user-attachments/files/19526517/koinly.csv)

pages/admin/pro/history.js
juddydev: Got it I'll updated this

juddydev:
Export CSV file for CoindLedger platform => SOLVED

@ihomp  could you check this?


https://github.com/user-attachments/assets/dc23421b-d312-4120-b048-f6b5cfb935f9

pages/admin/pro/history.js
```diff
@@ -252,16 +310,20 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
...
+        res.activities[i].receivedAmount = !sending ? res.activities[i].amountNumber : ""
         res.activities[i].receivedCurrency = !sending ? scvCurrency : ''
 
         res.activities[i].netWorthCurrency = selectedCurrency.toUpperCase()
 
         //sanitize memos for CSV
         res.activities[i].memo = res.activities[i].memo?.replace(/"/g, "'") || ''
+
+        // For CoinLedger platform
+        res.activities[i].coinLedgerTxType = res.activities[i].direction === 'sent' ? 'Withdrawal' : res.activities[i].direction === 'received' ? 'Deposit' : res.activities[i].txType
```
ihomp, line 325: we should not rely on "direction", it should rely on the "amount". 
1) if positive amount -> Deposit 
2) if negative balance -> Withdrawl 

Are there any other options available?
Because if only fee was paid it's not widtrawal. it's network fee, it should be an option for that. 

then if the negative balance is less or equal the fee - then it's "Fee" or equvalent

ihomp:
the "direction" should not be used for the balance changes types

pages/admin/pro/history.js
```diff
@@ -29,7 +29,9 @@ import Image from 'next/image'
 import { CSVLink } from 'react-csv'
 import DownloadIcon from '../../../public/images/download.svg'
 import { koinly } from '../../../utils/koinly'
-
+import { format } from 'date-fns'
```
ihomp, line 32: why do we need 'date-fns' in here? 
JS functions not enough to format a date?

ihomp: The project doesn't have date-fns installed, and I don't see the package is added in the package.json and yarn.lock 

it's nice sometimes to avoid installing new packages, and just replace them with a function.

pages/admin/pro/history.js
```diff
@@ -56,6 +58,26 @@ const showFiat = (fiat, selectedCurrency) => {
...
 
+const dateFormatters = {
+  Koinly: (timestamp) => {
+    // ISO format: YYYY-MM-DDTHH:MM:SS.000Z
+    return new Date(timestamp * 1000).toISOString();
+  },
+  CoinLedger: (timestamp) => {
+    // Format: MM/DD/YYYY HH:MM:SS in UTC
+    const date = new Date(timestamp * 1000);
+    return format(new Date(date.getTime() + date.getTimezoneOffset() * 60000), 'MM/dd/yyyy HH:mm:ss');
```
ihomp, line 69: would this work from unix seconds to format they need in utc? 

```
function formatUnixToUTC(unixSeconds) {
  const date = new Date(unixSeconds * 1000); // Convert to milliseconds

  const pad = (n) => n.toString().padStart(2, '0');

  const mm = pad(date.getUTCMonth() + 1);
  const dd = pad(date.getUTCDate());
  const yyyy = date.getUTCFullYear();

  const hh = pad(date.getUTCHours());
  const min = pad(date.getUTCMinutes());
  const ss = pad(date.getUTCSeconds());

  return `${mm}/${dd}/${yyyy} ${hh}:${min}:${ss}`;
}
```

pages/admin/pro/history.js
```diff
@@ -222,7 +280,7 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
         let sending = res.activities[i].amountInFiats?.[selectedCurrency]?.[0] === '-'
         res.activities[i].index = options?.marker ? activities.length + 1 + i : i + 1
         res.activities[i].amountExport = amountFormat(res.activities[i].amount)
-        res.activities[i].amountNumber = res.activities[i].amount?.value || res.activities[i].amount / 1000000
+        res.activities[i].amountNumber = res.activities[i].amount?.value || (parseFloat(res.activities[i].amount) / 1000000)
```
ihomp, line 283: didn't it work without parseFloat? it is string, but it's always integer

juddydev:
image

@ihomp  Are there any other options available?
Because if only fee was paid it's not widtrawal. it's network fee, it should be an option for that.

then if the negative balance is less or equal the fee - then it's "Fee" or equvalent


=> CoinLedger also support Fee Tx type

ihomp:
> => CoinLedger also support Fee Tx type

So, why did you remove the "Fee" option in your last commit?

pages/admin/pro/history.js
```diff
@@ -440,32 +514,44 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
...
-                    { label: 'Net Worth Amount', key: 'amountInFiats.' + selectedCurrency },
-                    { label: 'Net Worth Currency', key: 'netWorthCurrency' },
-                    { label: 'Label', key: '' },
-                    { label: 'Description', key: 'memo' },
-                    { label: 'TxHash', key: 'hash' }
-                  ]}
+                  data={processDataForExport(filteredActivities || [], platformCSVExport)}
+                  headers={
+                    platformCSVHeaders.find(
+                      (header) => header.platform.toLowerCase() === platformCSVExport.toLocaleLowerCase()
```
ihomp, line 539: toLocaleLowerCase? 
did you mean the toLowerCase?

ihomp (changes requested):

juddydev:
> > => CoinLedger also support Fee Tx type
> 
> So, why did you remove the "Fee" option in your last commit?

CoinLedger also support Fee tx type but It's only for Manual Input
But we need to import through CSV file so TX type "Fee" currently not supported. I have tried 
It just only support some these transaction type
image

So I think Fee also use Withdrawal TX Type
It's the best choice currently I think

pages/admin/pro/history.js
juddydev: I'll update this right now

ihomp (approved):
@Anna15170221 the code looks good for me now. 
If CSV imports works well for both platforms, we can merge it and give to some users to test.

PR #248, opened 2025-04-01
verified domains article

Anna15170221:
The line 51 should contain {DOMAIN}. I don't know how to include the curly braces {}; it causes an error


ihomp:
- the similar images for all the networks are better to keep in separate folders for each page, like 
/pages/verified-domains/checkmark

if the images are different for different networks, we have folders 
xrplexplorer/pages/verified-domains/checkmark
xahauexplorer/pages/verified-domains/checkmark

where verified-domains should be exactly the same as the route name (which is usually a page address)

That way we will keep the structure well maintained, and later it will be easier to automate things.

ihomp:
> The line 51 should contain {DOMAIN}. I don't know how to include the curly braces {}; it causes an error

Screenshot 2025-04-03 at 18 44 37

ihomp (changes requested):
- the images names and structure needs to be fixed

PR #253, opened 2025-04-15
fix: add cross mark to search bar

hiroshisekiya644:


ihomp:
This PR reverts the changes from v3 to v2. 
That cannot be accepted.

hiroshisekiya644:
I fix the problem and keep the current version.

ihomp:
Awesome, NOW it looks great! Thank you! 
Merging

Issue #219, opened 2025-02-20
can't delete taxon from filter


Anna15170221:
https://bithomp.com/en/nft-distribution
https://bithomp.com/en/nft-explorer
https://bithomp.com/en/nft-sales
When the issuer is removed from filter, the taxon can't be deleted - this is a bug and it should be fixed. 

https://github.com/user-attachments/assets/6cff016e-7d0e-4cf6-ad43-847c174e2d13

ihomp:
Closing it now.

PR #254, opened 2025-04-17
Bug fix of Issue #219

serendipity-seeker:
This is pull request for issue #219 .
Origin code was disabling taxon field unless issuer is not inputted or deleted.
So removed `disabled` field from taxon input.


ihomp:
That code breaks the old functionality, where the taxon would become disabled when there is no issue. 

Closing the PR, the issue is closed.

PR #255, opened 2025-04-17
Fix: can't delete taxon from filter (Issue #219)

pandablue0809:
## Description
Fixed issue where taxon input value wasn't clearing properly when issuer input was removed. Added key prop to FormInput component to force proper remounting and state reset.

## Changes Made
- Added `key={issuerInput || 'empty'}` to FormInput component to ensure proper reset of internal state when issuer changes


ihomp:
Nice code and nice description! I'll review it.

ihomp:
Awesome! 

どうもありがとう御座いました！

Merging and closing the issue!

Issue #213, opened 2025-02-17
#2 TokenTax CSV intergration


Anna15170221:
Instruction: https://help.tokentax.co/en/articles/1707630-create-a-manual-csv-report-of-your-transactions

PR #256, opened 2025-04-19
create export csv for token-tax

juddydev:
Create export csv for token-tax #213


ihomp:
@Anna15170221 can you please test this one?

Anna15170221:
Hi @juddydev 
could you please specify why the column **Exchange** is empty? 
In the ticket you will find the link where it's explained what is expected to be in each column.
If we are missing something in our API, as I already told before, you can ask the devs to add.
It's not fine to leave the empty fields without any reason, because the integration is not finalised in this case.

juddydev:
> Hi @juddydev could you please specify why the column **Exchange** is empty? In the ticket you will find the link where it's explained what is expected to be in each column. If we are missing something in our API, as I already told before, you can ask the devs to add. It's not fine to leave the empty fields without any reason, because the integration is not finalised in this case.

@ihomp @Anna15170221 
This column tells where/how the transaction took place. Put the exchange name if the trade was done over an exchange. If the trade relates to a deposit, withdrawal, or transaction originating from a wallet, put the wallet name (e.g. Exodus). If an over-the-counter trade, put OTC.

I can't see any fields for exchange in the API response

ihomp:
> > Hi @juddydev could you please specify why the column **Exchange** is empty? In the ticket you will find the link where it's explained what is expected to be in each column. If we are missing something in our API, as I already told before, you can ask the devs to add. It's not fine to leave the empty fields without any reason, because the integration is not finalised in this case.
> 
> @ihomp @Anna15170221 This column tells where/how the transaction took place. Put the exchange name if the trade was done over an exchange. If the trade relates to a deposit, withdrawal, or transaction originating from a wallet, put the wallet name (e.g. Exodus). If an over-the-counter trade, put OTC.
> 
> I can't see any fields for exchange in the API response

We can't have it in the API, as we don't know what platforms user uses, so the user can fill this manually in the token-tax. 

Do we need to have this fields and keept it empty, or should we just skip it and the export would work?

Anna15170221:
@ihomp @juddydev 
I think, we should keep all required fields in the CSV and leave empty the ones we can't fill.

ihomp (approved):

Issue #214, opened 2025-02-17
#3 ZenLedger CSV file intergration


Anna15170221:
Instruction: https://support.zenledger.io/en/articles/2615489-how-do-i-create-a-custom-csv

PR #257, opened 2025-04-19
create export csv for zenledger

juddydev:
Create export csv for zenledger #214


juddydev:
Because ZenLedger needs to determine the USD value of that coin, we need to add the USD value to it (some popular coins like XRP, BTC,... will automatically determine the USD value without adding) however, there will be some coins that do not have a value, meaning its USD value is 0, we must leave it empty, we cannot leave it 0 because when importing, there will be an error.

image

So some row have missing price (USD Value = 0)

ihomp:
We have a filter checkbox to discard small amounts so all entries with values around 0 won't be exported. 
The same filter with a specific value limit can be integrated for that export.

ihomp:
The PR is outdated; we need a new PR for the current version of the code. 
So, the function will be used for the time format, etc.

ihomp (changes requested):
We need a new PR, for the updated code base

pages/admin/pro/history.js
```diff
@@ -109,13 +116,33 @@ const processDataForExport = (activities, platform) => {
...
         : Math.abs(activity.amountNumber) <= activity.txFeeNumber
         ? 'Other Fee'
         : 'Deposit'
+    } else if (platform === 'ZenLedger') {
+      if (activity.amountNumber > 0) {
+        processedActivity.type = 'Receive'
+        processedActivity.receivedAmount = activity.amountNumber
+        processedActivity.receivedCurrency = activity.receivedCurrency
+
+        // USD Value
```
ihomp, line 132: USD value? 
Can they accept any fiat or do they need only USD? 

selectedCurrency it's fiat currency selected by user
but we can force it to be USD if needed.

ihomp (commented):
do they support all fiat currecnies? we have 40 of them.. or do they need only USD?

ihomp:
conflicts need to be resolved

juddydev:
> do they support all fiat currecnies? we have 40 of them.. or do they need only USD?

only supported USD

ihomp:
> > do they support all fiat currecnies? we have 40 of them.. or do they need only USD?
> 
> only supported USD

We have USD that can be used.

ihomp (approved):
Not tested, code looks ok

ihomp:
@Anna15170221 can you please test that PR?

Anna15170221:
@juddydev 
@ihomp I can't test 
Screenshot 2025-05-13 at 18 08 28

Anna15170221:
@juddydev @ihomp 
our CSV file is importing to the platform with multiple errors like "missing prices" and "insufficient balances".
This is not ok to leave it like this.
What can we do about it?
Screenshot 2025-05-14 at 13 14 41
Screenshot 2025-05-14 at 13 15 18
Screenshot 2025-05-14 at 13 15 42

ihomp:
what the threshold they have for "insufficient balances"? We can filter out all the amounts less than the minimum they need the same way as we do "remove dust transactions".

ihomp:
"missing prices" - is there usd values in the CSV? I guess they are missing

ihomp (changes requested):
Lets fix the formatting, so it will imported without errors

juddydev:
@ihomp 
In terms of missing prices error, the 'missing prices' error occurs because the USD value in the API response is returned as 0, indicating that this coin doesn't have a USD price available.

ihomp:
> @ihomp In terms of missing prices error, the 'missing prices' error occurs because the USD value in the API response is returned as 0, indicating that this coin doesn't have a USD price available.

What precision do we need here? I guess the real volume could be like 0.0000001 usd or so.. but most likely that can be ignored.. we need to ignore then dust transactions as we do for koinly.. we just need to know what is the minimum amount we need to report, so the rest will be ignored. 

otherwise we just need to chnage the presicion for usd values we return.

Anna15170221:
@juddydev @ihomp 
I can't test
Screenshot 2025-05-28 at 13 09 09

juddydev:
@ihomp cc @Anna15170221 
For this one, Missing price -> mean that ZenLedger platform cannot determine a fair market value so It cause Missing Price error, not on our side

will share the example
image
image
image

I have editted manually
I choose an Token ASDIR, add Value USD for it is 10 USD
Click save
but It's also Missing Price because Zenledger platform cannot determine the fair market value of this token so 10USD will be auto changed into 0 and cause Missing Price

ihomp:
Can't we send them the price in USD in our CSV Exports instead of manually adding the USD price on their website? 
Or do they reject the USD price we suggest in the CSVs?

juddydev:
> Can't we send them the price in USD in our CSV Exports instead of manually adding the USD price on their website? Or do they reject the USD price we suggest in the CSVs?

It has nothing to do with adding USD price or not, it depends on whether their platform (ZenLedger) can identify what token it is or not

ihomp:
@juddydev I guess we shouldn't see negative balances for XRP only transactions.

ihomp:
no activity for 2 weeks, shall we just drop this PR? @juddydev @Anna15170221  ?

ihomp:
ok, juddydev can reopen it later

Issue #215, opened 2025-02-17
#4 CoinTracking CSV file intergration


Anna15170221:
Instruction: If you want to create your own CSV file, please ensure the header format is like this:
"Type", "Buy Amount", "Buy Currency", "Sell Amount", "Sell Currency", "Fee", "Fee Currency", "Exchange", "Trade-Group", "Comment", "Date"
Optionally you can add those 4 columns at the end (after the "Date" column):
"Tx-ID", "Buy Value in Account Currency", "Sell Value in Account Currency", "Liquidity pool"
The Buy and Sell Value is the traded amount converted into your Account Currency (e.g. USD) at the time of the transaction. If you leave the Buy and Sell Values empty or set them to 0, they will be automatically calculated by CoinTracking.

PR #258, opened 2025-04-19
export csv for cointracking

juddydev:
Create export csv for CoinTracking #215


ihomp:
1. The similar code should be extracted as a function (done for timestamp)
2. The timestampExport was processed twice, initially and then through the function one more time 
3. The processDataForExport only included the timestamp, but not the other Export data. It's better to have them in the same place, as the function name suggests. 

@Anna15170221, can you load my last changes and test it after that?

Anna15170221:
The CSV file contains IDs instead of currencies. It was implemented only for Koinly, the other platforms shouldn't have them.

ihomp (approved):
fixes the issue, but its needs to be refactored

PR #260, opened 2025-04-21
Update 'no results' message to suggest adjusting filters

pandablue0809:
# Issue 
[Add text to NFT Explorer/NFT Sales pages](https://github.com/Bithomp/frontend-react/issues/208)

# Requrements

If there are no results but the filter is adjusted, in addition to "We couldn't find any NFTs with your search parameters." ADD: Try adjusting the filters to see results that match your search!

# Implemented
- The text was added/updated only on these pages:
  - [x] https://bithomp.com/en/nft-explorer
  - [x] https://bithomp.com/en/nft-sales


ihomp:
The issue states: `If there are no results but the filter is adjusted`

I would clarify it:
It doesn't matter if the filter was updated or not; if there are any present filters, then the message should incude "Try adjusting the filters to see results that match your search!", but if no filters on, we should have the old message without suggesting to change the parameters. 

So if all filters that narrow the results (even if they are on by default) are off, then the message should stay as it was. There is no NFTs.

ihomp (changes requested):
we shouldn't suggest change search parameters when there are no NFTs and there no affective the search filters

components/NftsComponent.js
```diff
@@ -406,7 +406,11 @@ export default function NftsComponent({
           setData([...nftsData, ...nftList])
         } else {
           if (marker === 'first') {
-            setErrorMessage(t('nfts.no-nfts'))
+            setErrorMessage(
+              hasActiveFilters() 
+                ? t('nfts.no-nfts')
+                : t('general.no-data')
```
ihomp, line 412: We didn't have "general.no-data" here before... 

so what we need here is 
```
setErrorMessage(t('nfts.no-nfts') + " " + (hasActiveFilters()  ? t('general.change-filters') : "" )) 
```

the nfts.no-nfts should stay as it used to be. "We couldn't find any NFTs with your search parameters."

we need to add one more 
`general.change-filters `

"Try adjusting the filters to see results that match your search!"

that whay when there is no nft and no filters it will be the same message as we had it, but when filters present - a new message to adjust them.

components/NftsComponent.js
```diff
@@ -611,6 +615,22 @@ export default function NftsComponent({
...
+  // Add this function to check if any filters are active
+  const hasActiveFilters = () => {
+    return !!(
+      issuer ||
+      taxon ||
+      owner ||
+      search ||
+      mintedPeriod && mintedPeriod !== 'all' ||
+      burnedPeriod ||
+      includeBurned ||
```
ihomp, line 627: should be `!includeBurned `
includeBurned - would return more results than without it.

components/NftsComponent.js
```diff
@@ -611,6 +615,22 @@ export default function NftsComponent({
...
+      includeWithoutMediaData ||
```
ihomp, line 628: includeWithoutMediaData - would return more results, we need to check that if it's not enabled. 
if !includeWithoutMediaData -> adjust filters..

components/NftsComponent.js
```diff
@@ -611,6 +615,22 @@ export default function NftsComponent({
...
+      (listTab === 'onSale')
+      // (listTab === 'nfts' && order !== 'mintedNew')
```
ihomp, line 630: order is sorting, not the filter, no need for sortings.. pls remove

pages/nft-sales.js
```diff
@@ -220,6 +220,21 @@ export default function NftSales({
...
   }
 
+  const hasActiveFilters = () => {
+    return !!(
+      issuer ||
+      taxon ||
+      buyer ||
+      seller ||
+      search ||
+      (period && period !== (xahauNetwork ? 'year' : 'week')) ||
```
ihomp, line 230: We need to check for the max period here, any period less than max - would narrow down the results.

pages/nft-sales.js
```diff
@@ -220,6 +220,21 @@ export default function NftSales({
...
+      (saleTab !== 'primaryAndSecondary') ||
+      includeWithoutMediaData ||
```
ihomp, line 232: `includeWithoutMediaData` would narrow down the search, we need to check here for `!includeWithoutMediaData`

pages/nft-sales.js
```diff
@@ -341,7 +356,11 @@ export default function NftSales({
           setSales([...salesData, ...newdata.sales])
         } else {
           if (marker === 'first') {
-            setErrorMessage(t('general.no-data'))
+            setErrorMessage(
+              hasActiveFilters() 
+                ? t('nfts.no-nfts')
+                : t('general.no-data')
```
ihomp, line 362: same thing as before

public/locales/de/common.json
```diff
@@ -624,7 +624,7 @@
     "search-by-owner": "Suche nach NFT Besitzer",
     "search-by-buyer": "Suche nach NFT Käufer",
     "search-by-seller": "Suche nach NFT Verkäufer",
-    "no-nfts": "Wir konnten keine NFTs mit Ihren Suchparametern finden.",
+    "no-nfts": "Wir konnten keine NFTs mit Ihren Suchparametern finden. Versuchen Sie, die Filter anzupassen, um Ergebnisse zu sehen, die Ihrer Suche entsprechen!",
```
ihomp, line 627: We need to have an additional instance, in `general`. rather then extending the same one.

ihomp (changes requested):
Check comments

components/NftsComponent.js
```diff
@@ -611,6 +613,21 @@ export default function NftsComponent({
...
   */
 
+  // Add this function to check if any filters are active
+  const hasActiveFilters = () => {
+    return !!(
+      issuer ||
+      taxon ||
+      owner ||
+      search ||
+      mintedPeriod && mintedPeriod !== 'all' ||
```
ihomp, line 623: I know that `&&` has priority over `|| `and it looks good, but in some places you add parentheses `()`, but not here; keeping the same style everywhere would be nice.

pages/nft-sales.js
```diff
@@ -220,6 +220,21 @@ export default function NftSales({
...
   }
 
+  const hasActiveFilters = () => {
+    return !!(
+      issuer ||
+      taxon ||
+      buyer ||
+      seller ||
+      search ||
+      (period && period !== 'all') ||
```
ihomp, line 230: here you have `()` :)

ihomp (approved):
Looks great.

Anna15170221:
@ihomp tested - works fine

Issue #217, opened 2025-02-17
#6 BlockPit CSV file integration


Anna15170221:
Instruction: https://help.blockpit.io/hc/en-us/articles/360011877920-How-to-import-my-CSV-Excel-history

PR #262, opened 2025-04-22
add export csv blockpit

juddydev:
BlockPit CSV integration #217

1. The Byte Order Mark (BOM), a Unicode character (U+FEFF) represented as EF BB BF in UTF-8
=> uFEFF=false is set for Blockpit to prevent adding a BOM (Byte Order Mark) to the CSV file, as Blockpit requires CSV files without BOM for successful import.
2. Example files: Both contain the same content, but only the second file can be imported into Blockpit.
[blockpit-csv-no-config-uFEFF.csv](https://github.com/user-attachments/files/19853896/blockpit-csv-no-config-uFEFF.csv)
[blockpit-uFEFF-false.csv](https://github.com/user-attachments/files/19853902/blockpit-uFEFF-false.csv)


ihomp (changes requested):
we need a new PR agains the updated codebase

juddydev:
With Blockpit, Could not parse timestamp -> because BlockPit has issues with these day
for example row data has time is: "13.04.2025 15:38:12" -> parse error but we change to "12.04.2025 15:38:12"
It will be successfull

pages/admin/pro/history.js
```diff
@@ -200,6 +211,22 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
           { label: 'Sell Value in Account Currency', key: '' },
           { label: 'Liquidity pool', key: '' }
         ]
+      },
+      {
+        platform: 'BlockPit',
+        headers: [
+          { label: 'Date (UTC)', key: 'timestampExport' },
+          { label: 'Integration Name', key: 'platform' },
+          { label: 'Label', key: 'blockPitTxType' },
```
ihomp, line 220: shouldn't the key be just `type` instead of `blockPitTxType`? 
on line 125 it's `type`

ihomp (changes requested):
- check how the type works here
- it's better to send PRs with the same changes which were sent in a different pr (`sending` replacement)

juddydev:
Before update the logic of blockPit. I have a question
For an example

```
{
"address": "rpeSQTwakWtoc2LJMx5ZTJBpbTHbtxxmcf",
"timestamp": 1738142602,
"ledgerIndex": 93782284,
"txIndex": 24,
"hash": "2B314D603E2C3E9522DBB189EFB6764AE8FC89FABDE6EB20EB3168033A1CF7E8",
"direction": "dex",
"txType": "OfferCreate",
"counterparty": null,
"st": null,
"dt": null,
"currency": "XRP",
"amount": "5",
"amountNative": "0.000005",
"amountInFiats": {
"usd": "0.00001550895"
},
"transferFee": null,
"transferFeeNative": null,
"transferFeeInFiats": {
"usd": 0
},
"txFee": "12",
"txFeeNative": "0.000012",
"txFeeInFiats": {
"usd": "0.00003722148"
},
"references": null,
"memo": null
},
```

This transaction is of type Fee, correct?
In the Blockpit template, if a transaction's type is Fee, we need to set receivedAmount and receivedCurrency to an empty string.

ihomp:
> Before update the logic of blockPit. I have a question For an example
> 
> ```
> {
> "address": "rpeSQTwakWtoc2LJMx5ZTJBpbTHbtxxmcf",
> "timestamp": 1738142602,
> "ledgerIndex": 93782284,
> "txIndex": 24,
> "hash": "2B314D603E2C3E9522DBB189EFB6764AE8FC89FABDE6EB20EB3168033A1CF7E8",
> "direction": "dex",
> "txType": "OfferCreate",
> "counterparty": null,
> "st": null,
> "dt": null,
> "currency": "XRP",
> "amount": "5",
> "amountNative": "0.000005",
> "amountInFiats": {
> "usd": "0.00001550895"
> },
> "transferFee": null,
> "transferFeeNative": null,
> "transferFeeInFiats": {
> "usd": 0
> },
> "txFee": "12",
> "txFeeNative": "0.000012",
> "txFeeInFiats": {
> "usd": "0.00003722148"
> },
> "references": null,
> "memo": null
> },
> ```
> 
> This transaction is of type Fee, correct? In the Blockpit template, if a transaction's type is Fee, we need to set receivedAmount and receivedCurrency to an empty string.

Here is the trasaction:
https://bithomp.com/en/tx/2B314D603E2C3E9522DBB189EFB6764AE8FC89FABDE6EB20EB3168033A1CF7E8

the address received 5 drops from selling on dex 1700 HGF tokens, but spent on Fee 12 drops 

So, Basically, user spent more on Fee that received from that trade..

Therefore such transactions where (The FEE is higher than what was received) can be counted as FEE transaction.

juddydev:
> > Before update the logic of blockPit. I have a question For an example
> > ```
> > {
> > "address": "rpeSQTwakWtoc2LJMx5ZTJBpbTHbtxxmcf",
> > "timestamp": 1738142602,
> > "ledgerIndex": 93782284,
> > "txIndex": 24,
> > "hash": "2B314D603E2C3E9522DBB189EFB6764AE8FC89FABDE6EB20EB3168033A1CF7E8",
> > "direction": "dex",
> > "txType": "OfferCreate",
> > "counterparty": null,
> > "st": null,
> > "dt": null,
> > "currency": "XRP",
> > "amount": "5",
> > "amountNative": "0.000005",
> > "amountInFiats": {
> > "usd": "0.00001550895"
> > },
> > "transferFee": null,
> > "transferFeeNative": null,
> > "transferFeeInFiats": {
> > "usd": 0
> > },
> > "txFee": "12",
> > "txFeeNative": "0.000012",
> > "txFeeInFiats": {
> > "usd": "0.00003722148"
> > },
> > "references": null,
> > "memo": null
> > },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This transaction is of type Fee, correct? In the Blockpit template, if a transaction's type is Fee, we need to set receivedAmount and receivedCurrency to an empty string.
> 
> Here is the trasaction: https://bithomp.com/en/tx/2B314D603E2C3E9522DBB189EFB6764AE8FC89FABDE6EB20EB3168033A1CF7E8
> 
> the address received 5 drops from selling on dex 1700 HGF tokens, but spent on Fee 12 drops
> 
> So, Basically, user spent more on Fee that received from that trade..
> 
> Therefore such transactions where (The FEE is higher than what was received) can be counted as FEE transaction.

thanks, ok please check this pr now

ihomp:
@juddydev can the " (optional)" be removed from CSV labels? 

@Anna15170221, can you pls test this PR?

juddydev:
@ihomp 
We cannot remove (optional) label
It will cause the error "Template not valid"

Anna15170221:
@ihomp 
works fine and CSVs are successfully imported to the platform.

PR #263, opened 2025-04-22
create export csv for cryptotax

juddydev:
1. Example output:
[crypto-tax-csv.csv](https://github.com/user-attachments/files/19856012/crypto-tax-csv.csv)

2. Test import
image


ihomp (changes requested):
We need a new PR agains the latest code base

pages/admin/pro/history.js
```diff
@@ -116,6 +121,12 @@ const processDataForExport = (activities, platform) => {
         : Math.abs(activity.amountNumber) <= activity.txFeeNumber
         ? 'Other Fee'
         : 'Deposit'
+    } else if (platform === 'CryptoTax') {
+      processedActivity.cryptoTaxTxType = !sending
```
ihomp, line 125: Shouldn't it be here processedActivity.type instead of processedActivity.cryptoTaxTxType? 

because below you use `type` in headers 
And that would also be good for consistency

pages/admin/pro/history.js
```diff
@@ -200,6 +211,26 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
           { label: 'Sell Value in Account Currency', key: '' },
           { label: 'Liquidity pool', key: '' }
         ]
+      },
+      {
+        platform: 'CryptoTax',
+        headers: [
+          { label: 'Timestamp (UTC)', key: 'timestampExport' },
+          { label: 'Type', key: 'type' },
```
ihomp, line 219: it's `type` here

pages/admin/pro/history.js
```diff
@@ -375,6 +406,23 @@ export default function History({ queryAddress, selectedCurrency, setSelectedCur
 
         //sanitize memos for CSV
         res.activities[i].memo = res.activities[i].memo?.replace(/"/g, "'") || ''
+
+        // For CryptoTax platform
```
ihomp, line 410: As it's specific for a platform, we shouldn't do it for everyone. 
The code needs to be moved into the `processDataForExport` function

ihomp (changes requested):
Requires fixes

ihomp (approved):
The code looks ok, but we need a test

ihomp:
@juddydev can the " (Optional)" be removed from the CSV Labels? 

@Anna15170221 Can you please test this PR?

juddydev:
@ihomp 
We cannot remove (optional) label
It will cause the error "Template not valid"

Anna15170221:
@ihomp 
our CSVs are successfully imported to the platform

PR #268, opened 2025-04-29
whales

Anna15170221:
SEO component, short text, sitemap


ihomp:
dropping this PR
- sitemap already merged 
- we should not add SEO into the component, it should be on the pages.

Issue #133, opened 2024-05-27
NFT mint (XRPL)


ihomp:
by chrische

```
Bidirectional pointer for NFTs on XRPL
It would be nice to get the precalculated NFTokenID before the actual minting. Then I could add this information in the metadata file and on the image (exif) in the form of {chain:NetworkID:NFTokenID} to ensure a closed information loop. (bidirectional pointer).
```

ihomp:
Here are the docs for nftokenMint:
https://xrpl.org/docs/references/protocol/transactions/types/nftokenmint

for a simple first version mint, we just need a few fields:

TransferFee - optional for a user to add
Issuer - address search filed (like in the NFT explorer filters), option for user
URI - optional (same as in NFT mint on Xahau)
NFTokenTaxon - optional for a user (same verification as in nft-explorer search)
Amount - optional for user, as for the first simple version, its ok to use only XRP, later we can update it to use IOUs (tokens)
Expiration - choose date and time element - optional for a user
Destination - address search filed (like in the NFT explorer filters) - optional for a user

Flags
- Burnable
- onlyXRP
- Transferable
- Mutable 

As the first simple version, no translation needs to be added; only English is ok.

PR #269, opened 2025-04-30
NFT mint (XRPL)

pandablue0809:
# Issue

[NFT mint (XRPL)](https://github.com/Bithomp/frontend-react/issues/133)

# Discription
XRPL-NFT-Mint-04-30-2025_11_45_PM


ihomp:
- the tabs [XRPL | Xahau] looks good, but it's missing all the test/dev networks 
For example, on tha page https://xahauexplorer.com/en/faucet - we have listed all the networks 

We have the network change in the top right corner. 
so, I'm not sure if we should add all the test networks in the tabs as well, or shall we just remove the tabs... as there is a switcher in the right corner. 

What are your thoughts on this? @pandablue0809 @Anna15170221

ihomp:
@pandablue0809 I see you have "Add digest" for the XRPL. I thought Digest was only available for uritokens on Xahau. Does it do anything for the XRPL transaction? if not, that can be removed from XRPL.

ihomp:
`Issuer (optional - if different from your account):`

That might look a bit confusing for newcomers. 
From UI it would be better to do it in the way we did it for Digest in Xahau. 

a checkbox like 
**Mint on behalf of another account**

And when it's checked, show the field 
**Issuer**

ihomp:
I see Expiration on the screenshot, but I don't see it the page now. 
It's ok to do it without Expiration at the moment; the expiration can be added later in a separate PR.

ihomp:
I think we also need here a checkbox like

**Create a Sell offer** 
When checked, then three new fields will be shown:

- Amount
- Destination
- Expiration (it's ok to have it in a separate PR later)

pandablue0809:
hi @ihomp 
I fixed
plz check


> * the tabs [XRPL | Xahau] looks good, but it's missing all the test/dev networks
>   For example, on tha page https://xahauexplorer.com/en/faucet - we have listed all the networks
> 
> We have the network change in the top right corner. so, I'm not sure if we should add all the test networks in the tabs as well, or shall we just remove the tabs... as there is a switcher in the right corner.
> 
> What are your thoughts on this? @pandablue0809 @Anna15170221

pandablue0809:
> I think we also need here a checkbox like
> 
> **Create a Sell offer** When checked, then three new fields will be shown:
> 
> * Amount
> * Destination
> * Expiration (it's ok to have it in a separate PR later)
> 
> Thi all Sell offer functionality can be added in a separate PR later, if you want to make the first PR simpler.

pandablue0809:
hi @ihomp 
I fixed
About Expiration  and Create a Sell offer I will implement in next PR
Plz check

ihomp:
@pandablue0809 

The created component `NftMintTabs` is basically a copy of the existing one, `FaucetTabs`
It would be nice to reuse the same component for easier maintenance. 

- remove NftMintTabs
- rename FaucetTabs to NetwortkTabs (do a bit of renaming inside of the component)
- replace the use of FaucetTabs and NftMintTabs to NetwortkTabs

ihomp:
> hi @ihomp I fixed About Expiration and Create a Sell offer I will implement in next PR Plz check

Oh, I see, the Expiration shows up when someone enters the Amount. 

So we have two options for the PR: 

1) To make the first PR simple, you can hide Offer Creation fields, such as 
- Destination 
- Amount
- Expiration
So, they can be enabled in the next PR. 

OR

2) Add a checkbox **Create a Sell offer** 

That would show those additional fields
- Destination 
- Amount
- Expiration

ihomp:
From a UI perspective:

checkbox: **Create an initial Sell offer**:
Amount: (Initial listing price in XRP)
Expiration: (How long your offer is valid for)
Destination: (Recipient of NFT, if you use a marketplace as a broker, or Private offers)

ihomp:
It would also be good to split the flags checkboxes from the terms checkboxes. Maybe add some space in between?

ihomp:
for the fields with addresses, `Issuer` and `Destination` for the <addressInput/> component, we need to use `hideButton={true}` to hide the search button. 

Also the placeholder texts should be as simple as 
Issuer address 
Destination address

We need to make sure that when a username is entered, the address is passed to the transaction, not the username.

pandablue0809:
hi @ihomp 
I fixed
I added  Create a Sell offer in this PR
Plz review again

ihomp:
`NFT Taxon (classification number, required):`
Lets update to smth like 
`NFT Taxon (Collection identifier. Leave as 0 for the issuer’s first collection).`

ihomp:
Screenshot 2025-05-06 at 10 35 02

- I can not enable the transferable flag if it was checked out, and then I've entered the transfer fee.
- There is no error, when there is a transfer fee and i'm trying to click on tranfarable flag - it's just doesn't allow to uncheck it
- The error that the flag is required somehow under the mutable flag.. I guess it's better to place it near the transfarable flag.

ihomp:
Fail NFTMintXahau has some probems: 
Screenshot 2025-05-06 at 10 41 11

ihomp:
There are also problems on the NftMontXRPL 
Screenshot 2025-05-06 at 13 17 56

ihomp:
The components should be named:
**NFTokenMint** - the one you call now NftMintXRPL
**URITokenMint** - the one you call now NftMintXAHAU

as they can both exist on both XRPL and Xahau as amendments.

ihomp:
`mb4` class was added 14th times, but I don't think it was defined.. where it comes from?

ihomp:
on XRPL you can create a sell offer when you mint. in one transaction, you don't need to send a second transaction to create it separately.  

so, we don't need here
`TransactionType: 'NFTokenCreateOffer',`

It's simple: when the user clicks Create Sell Offer, then the `NFTokenMint` transaction should also include Amount, and optionally (Destination and Expiration), as described in docs

https://xrpl.org/docs/references/protocol/transactions/types/nftokenmint

we don't need a second transaction here, it can be done in one

ihomp:
When Issuer is specified, it is not added into the transaction...

pandablue0809:
hi @ihomp 
I fixed
plz review again

ihomp:
My suggestion is 
- to Remove the "NFT Flags" text 

Make a new order for fields:
- URI
- NFT Taxon
- checkbox Transferable (checked by default)
- Transfer fee field (hide that field if there is no checkbox in the Transferable)
We can write there 0 as default value, but we do not need to send the transfer fee field to transaction if it is 0. 
We can change the "Transfer Fee" to "Royalty (paid to the issuer, 0-50%)"
- Mutable 
- Only XRP
- Burnable 
- Create Sell Offer 
- Mint of behalf of a different account

ihomp:
`<SEO /> ` can not be on the page and inside of the each component. That will include it twice on each page. 

We can keep the SEO only on the page level: `nft-mint.js`, but depending on `xahauNetwork` variable to have different strings for title and description.

ihomp:
It seems some of my code was deleted from the URITokenMint, that will brake the functionality we used to have 

The fileds could be specified in the URI, so the users could find easier time to mint NFTs. 

Here is the code I'm talking about:


```
export const getServerSideProps = async (context) => {
  const { query, locale } = context
  const { uri, digest } = query
  return {
    props: {
      uriQuery: uri || '',
      digestQuery: digest || '',
      isSsrMobile: getIsSsrMobile(context),
      ...(await serverSideTranslations(locale, ['common']))
    }
  }
}
```

here we were taking `uri` and `digest` get params from the page 

`const [uriValidDigest, setUriValidDigest] = useState(isIdValid(digestQuery))`
here it was set


From the place where we take v2/metadata, the console.log was removed in case of an error 

```
.catch((error) => {
        console.log(error)
        setMetadataStatus('error')
      })
```

Please do not remove the comment in the code we have, such as 

```
if (Date.now() - startTime < 120000) {
          // 2 minutes
```

why this was removed?

```
useEffect(() => {
    if (agreeToSiteTerms || agreeToPrivacyPolicy) {
      setErrorMessage('')
    }
  }, [agreeToSiteTerms, agreeToPrivacyPolicy])

  useEffect(() => {
    if (calculateDigest) {
      setDigest('')
    }
  }, [calculateDigest])


  useEffect(() => {
    let queryAddList = []
    let queryRemoveList = []
    if (digest) {
      queryAddList.push({
        name: 'digest',
        value: digest
      })
      setErrorMessage('')
    } else {
      queryRemoveList.push('digest')
    }

    if (uri) {
      queryAddList.push({
        name: 'uri',
        value: uri
      })
      setErrorMessage('')
    } else {
      queryRemoveList.push('uri')
    }

  }, [digest, uri])
```

The whole functionality to deal with get params was removed... 
That's needs to be returned; please do not remove the features.

ihomp:
There are copy-pasted things in NFTokenMint from URITokenMint, that are not in use.

There are two options:
1) Add the button "load metadata" and a field to show the loaded metadata
OR
2) Remove the unused copy-pasted functionality: getMetadata, checkDigest, setMetadataStatus, update, interval, etc which are not in use

Copy-pasting in general is not nice, the code would be better extracted as a component and functions, so they will be reused in both places. that is for easier maintaining of the code.

pandablue0809:
hi @ihomp 
I fixed
Plz check

ihomp:
I see you removed one of your text fixes:
"The NFT was sucefully minted" you had it as "The NFT was successfully minted"

ihomp:
I see you use a class `mb-2`, I don't see it's defined anywhere. Where does it come from?

ihomp:
@Anna15170221 can you test that uritoken mint on xahau works as before? 
it should add the uri and digest in the URI and also it should read them right away if they are in the URI.

ihomp:
@pandablue0809 

in the `nft-mint.js `

I see that you pass two variables `account` and `refreshPage`

```
            <NFTokenMint
              account={account} 
              setSignRequest={setSignRequest}
              refreshPage={refreshPage}
            />
```

but they are not in use in the `NFTokenMint` component 
`export default function NFTokenMint ({ setSignRequest })`

so, you might be able to remove them from variables passed to `NFTokenMint` and from 
`export default function NftMint({ account, setSignRequest, refreshPage, uriQuery, digestQuery }) {`

pandablue0809:
> I see you use a class `mb-2` 12 times, I don't see it's defined anywhere. Where does it come from?

sorry, it's my mistake

ihomp:
In the directory `components`, we are trying to have the same subdirectory structure as in the directory `pages`.

So the two new components, `NFTokenMint` and `URITokenMint`, should be inside the 
`components/services/nft-mint/` directory. 

The directory `components/nft/` should only have components related to the page `pages/nft`

ihomp:
Instead of `!taxon || isNaN(parseInt(taxon))`, it's better to reuse the function we already have to check the Taxon Validity

The function called `isValidTaxon` in the `utils.js `

Our function also checks the max value.

ihomp:
```
parseInt(transferFee) > 50000)) {
      setErrorMessage('Transfer Fee must be between 0 and 50000 (0-50%)')
```

Should you check here from 0 to 50, as that's what users enter in the field? 

`parseInt(transferFee) > 50` ? 

`Transfer Fee must be between 0% and 50%`

ihomp:
Not an issue, but I see things like 
`<p >` instead of `<p>`
It's nice to use the prettier extension to fix formatting.
We have prettier settings specified in the file `.prettierrc` in the root directory; which should be automatically taken if you use the prettier extension in your code editor.

ihomp:
Amount `0` - shouldn't give an error, it's a valid amount

ihomp:
Expiration should be in the Ripple epoch, as in the specification. 
So, instead of 
`mintRequest.Expiration = Math.floor(Date.now() / 1000) + expiration * 24 * 60 * 60`
it should be
`mintRequest.Expiration = Math.floor(Date.now() / 1000) + expiration * 24 * 60 * 60 - 946684800`

ihomp:
> > I see you use a class `mb-2` 12 times, I don't see it's defined anywhere. Where does it come from?
> 
> sorry, it's my mistake

Just remove them, as they seem unnecessary.
I also found `mb-4`, `mt-4`, `mt-6`

ihomp:
Component `NFTokenMint.js` is missing `server` that should be imported
`import { encode, server } from '../../utils' `

It is used on the line 420 
`{server}/nft/{minted}`

ihomp (changes requested):
requires some changes, as specified in comments

ihomp:
`Mutable` flag only works on the devnet for now. 
so we need to show only when 
`network === 'devnet'`
where `network` is from `utils.js`

ihomp:
Just as a cleanup:

We can remove:

```
      Memos: [
        {
          Memo: {
            MemoData: encode('NFT Mint')
          }
        }
      ]
```

from both requests in NFTokenMint and in URITokenMint

pandablue0809:
hi @ihomp 
I fixed
plz check

ihomp:
I've updated paths for components and utils.. otherwise I couldn't test =)

pandablue0809:
hi @ihomp 
I fixed

ihomp (approved):
Looks good!

Anna15170221:
@ihomp 
I've tested point 1 - works fine.
Testnet and Devnet are not currently working well (seems the reason is not from our side - can't sign the transactions, tried with different wallets) - will check later. 
JFYI, I've minted nft on the mainnet, expiration date is shown in nft history, but not in transaction itself.
https://bithomp.com/en/nft/0008000049EB0C5CBFBD4CD6558FBE6BCD383F47DB2E3B55491CED2005094E85

ihomp:
@Anna15170221 works on devnet for me with CrossMark

ihomp:
@pandablue0809 

when user enters the Amount as 0
it should be added to the transaction
`Amount: "0"`

I don't think this is the case now.

ihomp:
We should not add to tx:

Flags: 0

If there are no Flags, do not add Flags to the transaction.

Anna15170221:
@ihomp 
tested Devnet with Gem - worked fine now
https://dev.xrplexplorer.com/en/nft/001827108DCD674725CDA3A730E90CC197647CA8B6C396345236D64800298FAD
and Testnet with Gem
https://test.xrplexplorer.com/nft/000800008DCD674725CDA3A730E90CC197647CA8B6C396343894268F0020E8F4

ihomp:
I'll merge this, but we need fixes for flags:0 and Amount:0

PR #270, opened 2025-05-03
update export csv coinledger and cointracking

juddydev:


ihomp:
@juddydev, what would this PR change? Doesn't the code on lines 111-119 do the same thing?

juddydev:
> @juddydev, what would this PR change? Doesn't the code on lines 111-119 do the same thing?

Yes It do the same, I remove duplicate code, we already checked sending before, just use this variable again

Issue #204, opened 2025-01-10
We need to add the cross mark to the search bar 


Anna15170221:
This is the page, which should be updated: https://bithomp.com/en/account/
We have two issues here and we are expecting to get two separate pull requests.
1. Please add a cross mark"x" in the search bar to quickly clear the field. The same is already done here in the filter: https://bithomp.com/en/nft-explorer
2. When we enter a name or address into the search bar and open someone's account page, only the address should remain in the search bar when the page is open. The username or service name should appear above the search bar. **It should look the following way:** 

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/3eb8b698-69dc-4e58-bd92-1a2d2553ac64" />

**AND NOT LIKE THIS:**

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/a39322e8-ad88-42f3-a009-abfdb637b498" />

**OR NOT LIKE THIS:**

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/98ef2e6f-1aff-47bc-87d4-7404369ee6c6" />

ihomp:
@Anna15170221, please close this issue and open a small new one about the added crossmark that is not working on mobile.

PR #272, opened 2025-05-05
We need to add the cross mark to the search bar 

pandablue0809:
# Issue 

[We need to add the cross mark to the search bar ](https://github.com/Bithomp/frontend-react/issues/204)

# Screenshot
[Screencast from 05-06-25 01:33:24.webm](https://github.com/user-attachments/assets/4b30a041-8b7f-4cc6-9d12-515cdc87ddff)


pages/account/[[...id]].js
```diff
@@ -165,10 +165,7 @@ export default function Account({
 
   useEffect(() => {
     if (!selectedCurrency) return
-    if (data?.address) {
-      checkApi({ noCache: true })
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+    checkApi({ noCache: true })
```
Comment (ihomp) [inline pages/account/[[...id]].js:168]: That will send the additional request to the backend from the client side and that will trigger the re-render of the whole page - fix fixes the issue. 

- we don't need to send a new request, as all data we need has already been received. 
- no need to touch this function, as it's to specifically to update the page after submitting a transaction (no cache - as we need the latest version), and for the times when historical ledgers are checked. (ledgerTimestamp)
That function is used to FORCE to UPDATE the data for already loaded accounts, so please return it as it used to be.

ihomp (changes requested):
No need to send an additional request.

pandablue0809:
hi @ihomp 
plz check
I fixed

pages/account/[[...id]].js
```diff
@@ -168,7 +172,6 @@ export default function Account({
     if (data?.address) {
       checkApi({ noCache: true })
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Comment (ihomp) [inline pages/account/[[...id]].js:171]: As you removed this line, now I'm getting an error in VSCode:

`React Hook useEffect has missing dependencies: 'checkApi', 'data?.address', and 'selectedCurrency'. Either include them or remove the dependency array.`

ihomp (changes requested):
resolve the warning

ihomp (approved):
That looks good

Issue #276, opened 2025-05-09
X Mark on account page still doesn't work well.


Anna15170221:
@pandablue0809 
The x mark works well only for the first search, for the further searches it still doesn't clear the field (watch the video).
Please fix.
https://bithomp.com/en/account/
https://github.com/user-attachments/assets/972a78fc-40d5-4f93-ab5a-62540e20f8b7

0xdavinci:
@Anna15170221 I reviewed your video. My code already fixes the issue. I'm curious why you haven't used my code, as I've made a genuine contribution to your project.

ihomp:
> [@Anna15170221](https://github.com/Anna15170221) I reviewed your video. My code already fixes the issue. I'm curious why you haven't used my code, as I've made a genuine contribution to your project.

- Thank you for your contribution, but we already had another dev assigned for the issue; we didn't ask you to solve it. 
- Your solution wasn't as clear as the one we accepted; your PR included unnecessary changes. 
- Your solution also didn't solve the issue with the second search. 

We haven't expected any contributions from you, and we do not expect any new ones.

PR #277, opened 2025-05-09
X Mark on account page still doesn't work well.

pandablue0809:
# Issue
[X Mark on account page still doesn't work well.](https://github.com/Bithomp/frontend-react/issues/276#issuecomment-2866235143)
# Description
Hide the selected value by setting the .react-select__single-value opacity to 0.


ihomp:
@Anna15170221 can you test pls?

ihomp:
Wouldn't it be possible to clear the field and the search rather than making it invisible? 
That way we would see again the placeholder "Enter an address ... etc."

pandablue0809:
hi @ihomp 
I fixed
plz review again

Anna15170221:
@pandablue0809 
it got better again. 
But we still see the drop-down list with the previous search results when the field is already cleared (when searching for the second time and further). I assume, it should not appear.

https://github.com/user-attachments/assets/58a6c9e2-0e67-431f-bacf-93785d0da8f5


@ihomp do you agree?

pandablue0809:
hi @Anna15170221 
I fixed. plz check

Anna15170221:
@pandablue0809 
it works fine now for both mobile and desktop. Thank you.
@ihomp kindly review

ihomp:
I see the `searchOnFocus` is commented. 

It's OK to delete it.

ihomp (changes requested):
Please remove the commented part ;)

pandablue0809:
hi @ihomp 
I fixed
plz check

ihomp:
@Anna15170221 if the issues is solved, please close the related ticket

PR #278, opened 2025-05-12
update ui choose platform on mobile

juddydev:
image


Anna15170221:
@ihomp 
works on mobile now

Issue #280, opened 2025-05-13
XRPL NFT Mint page/ corrections are required


Anna15170221:
https://bithomp.com/en/services/nft-mint
1. We need the uri and taxon fields to be automatically filled in if they are passed via GET (the same way we currently handle uri and digest on Xahau).

2. When a user enters values into the taxon or uri fields, those values should be reflected in the URL — just like we do on Xahau.

ihomp:
@Anna15170221, please test locally on the main branch and close (not deployed yet)

Anna15170221:
tested, works fine

PR #283, opened 2025-05-14
XRPL NFT Mint page/ corrections are required

pandablue0809:
# Issue
[XRPL NFT Mint page/ corrections are required](https://github.com/Bithomp/frontend-react/issues/280)
# Description
-uri and taxon fields to be automatically filled in if they are passed via GET
-When a user enters values into the taxon or uri fields, those values should be reflected in the URL — just like we do on Xahau


ihomp:
1. you can crate a new branch, to keep the hsitory clear.
You just delete this branch, and then fork the latest **main** branch, and then code on top of it. 

2. we already have solutions to work with Get Params in xahau nft mint, it's better to reuse them. 

`nft-mint.js`
`const { uri, digest } = query`
you can just taxon here
const { uri, digest, taxon } = query

and pass it to a component 
`taxonQuery: taxon || '',`

`<NFTokenMint uriQuery={uriQuery} taxonQuery={taxonQuery}  />`

then in the NFTokenMint component we need smth like that 

```
  useEffect(() => {
    let queryAddList = []
    let queryRemoveList = []
    if (taxon !== undefined) {
      queryAddList.push({
        name: 'taxon',
        value: taxon
      })
      setErrorMessage('')
    } else {
      queryRemoveList.push('taxon')
    }
    if (uri) {
      queryAddList.push({
        name: 'uri',
        value: uri
      })
      setErrorMessage('')
    } else {
      queryRemoveList.push('uri')
    }
  }, [taxon, uri])
```

similiar to what we have in URITokenMint

ihomp:
we have a function to check taxon validity, we need to re-use it. 
`isValidTaxon` in utils

ihomp:
So, no new helpers are needed, and it's easier to verify and read the code.

pandablue0809:
hi @ihomp 
I fixed
plz check

components/Services/NftMint/NFTokenMint.js
```diff
@@ -1,18 +1,20 @@
...
-  const [uri, setUri] = useState('')
+export default function NFTokenMint({ setSignRequest, uriQuery , taxonQuery  }) {
+  const router = useRouter()
+  const [uri, setUri] = useState(uriQuery)
   const [agreeToSiteTerms, setAgreeToSiteTerms] = useState(false)
   const [agreeToPrivacyPolicy, setAgreeToPrivacyPolicy] = useState(false)
   const [errorMessage, setErrorMessage] = useState('')
   const [minted, setMinted] = useState('')
-  const [taxon, setTaxon] = useState('0')
+  const [taxon, setTaxon] = useState(taxonQuery)
```
ihomp, line 17: it's rather here should be `taxonQuery || '0'`

or taxonQuery should be as '0' by default when it's not in query params

components/Services/NftMint/NFTokenMint.js
```diff
@@ -165,6 +187,7 @@ export default function NFTokenMint({ setSignRequest }) {
     }
 
     setSignRequest({
+      redirect: 'nft',
```
ihomp, line 190: why would you add redirect here?

components/Services/NftMint/NFTokenMint.js
```diff
@@ -180,7 +203,7 @@ export default function NFTokenMint({ setSignRequest }) {
         {!minted && (
           <>
             {/* URI */}
-            <span className="input-title">URI that points to the data or metadata associated with the NFT:</span>
+            <p>URI that points to the data or metadata associated with the NFT:</p>
```
ihomp, line 206: why do you chnage it back?

pages/services/nft-mint.js
```diff
@@ -9,18 +9,19 @@ import NFTokenMint from '../../components/Services/NftMint/NFTokenMint'
 
 export const getServerSideProps = async (context) => {
   const { query, locale } = context
-  const { uri, digest } = query
+  const { uri, digest, taxon } = query
   return {
     props: {
       uriQuery: uri || '',
       digestQuery: digest || '',
+      taxonQuery: taxon || '',
```
ihomp, line 17: here it could be `'0' `inatead of `''` for default

ihomp (changes requested):
do not change <inputs> back to <p>

components/Services/NftMint/NFTokenMint.js
pandablue0809: I tried to do the same as URITokenMint.
back to the original?

pandablue0809:
@ihomp 
I fixed
plz check

components/Services/NftMint/NFTokenMint.js
pandablue0809: I already update, plz fix

ihomp (changes requested):
Please return `<br/><span>` instead of `<p>  `

components/Services/NftMint/NFTokenMint.js
```diff
@@ -273,8 +291,6 @@ export default function NFTokenMint({ setSignRequest }) {
               </CheckBox>
             </div>
 
-            <br />
```
ihomp, line 276: please return it too

pandablue0809:
I fixed @ihomp

components/Services/NftMint/NFTokenMint.js
```diff
@@ -302,14 +324,17 @@ export default function NFTokenMint({ setSignRequest }) {
...
+                  <AddressInput
+                    title="Destination (optional - account to receive the NFT):"
+                    placeholder="Destination address"
+                    setValue={onDestinationChange}
+                    initialValue={destination}
+                    name="destination"
+                    hideButton={true}
+                  />
+                </div>
+                <br />
```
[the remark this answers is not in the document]
pandablue0809: hi @ihomp 
<br /> is needed in here

Issue #281, opened 2025-05-14
New Page – Account Settings Update


Anna15170221:
We need to create a new page where the users can change their account settings.

- All fields listed in the specifications should be included:
https://xrpl.org/docs/tutorials/how-tos/manage-account-settings
https://xrpl.org/docs/references/protocol/transactions/types/accountset#accountset-flags

- The current account settings must be fetched from our API based on the user's address.
- The page should display the current settings exactly as they are for the user.
- The user should be able to modify these settings and submit changes.

ihomp:
The docs for `v2/address` are here:
https://docs.bithomp.com/#get-information-for-an-address

you would need `ledgerInfo=true`

ihomp:
There are 3 steps for the ticket:
1) Add flags 
https://xrpl.org/docs/references/protocol/transactions/types/accountset#accountset-flags

asf flags can be enabled only one by one
tfFlags can be enabled at ones - in one transaction

Flags that should be enabled/disabled in one transaction:

tfRequireDestTag | 0x00010000 | 65536 | asfRequireDest (SetFlag) | Require a destination tag to send transactions to this account.
-- | -- | -- | -- | --
tfOptionalDestTag | 0x00020000 | 131072 | asfRequireDest (ClearFlag) | - **REMOVING** tfRequireDestTag
tfRequireAuth | 0x00040000 | 262144 | asfRequireAuth (SetFlag) | Require authorization for users to hold balances issued by this address. Can only be enabled if the address has no trust lines connected to it.

tfOptionalAuth | 0x00080000 | 524288 | asfRequireAuth (ClearFlag) | **REMOVING** tfRequireAuth
tfDisallowXRP | 0x00100000 | 1048576 | asfDisallowXRP (SetFlag) | XRP should not be sent to this account. (Advisory; not enforced by the XRP Ledger protocol.)
tfAllowXRP | 0x00200000 | 2097152 | asfDisallowXRP (ClearFlag) | **REMOVING** tfDisallowXRP

Flags that can be enabled one by one only:


asfAccountTxnID | 5 | (None) | Track the ID of this account's most recent transaction. Required for AccountTxnID
-- | -- | -- | --
[...]

PR #286, opened 2025-05-19
Feat/281/account setting page

pandablue0809:
# Issue
[New Page – Account Settings Update](https://github.com/Bithomp/frontend-react/issues/281)

# Description

image


components/Layout/Header/index.js
```diff
@@ -432,6 +432,7 @@ export default function Header({
                 <Link href={'/account/' + address}>{t('signin.actions.view')}</Link>
                 <a href={server + '/explorer/' + address}>{t('signin.actions.my-transactions')}</a>
                 <Link href={'/nfts/' + address}>{t('signin.actions.my-nfts')}</Link>
+                <Link href="/account-settings/">{t('signin.actions.account-settings')}</Link>
```
ihomp, line 435: This link will be enabled for all the networks, including Xahau. It's ok though, but we need to make sure it works on xahau.

pages/account-settings.js
```diff
@@ -0,0 +1,252 @@
...
+  const { locale } = context
+  return {
+    props: {
+      isSsrMobile: getIsSsrMobile(context),
+      ...(await serverSideTranslations(locale, ['common', 'account-settings']))
+    }
+  }
+}
+
+// XRPL AccountSet flag values mapping
```
ihomp, line 20: Ok, this is for XRPL, but Xahau is must be similiar, they probably jsut miss the `disallowIncomingNFTokenOffer`.  

we need one for xahau too then.

pages/account-settings.js
```diff
@@ -0,0 +1,252 @@
...
+        const tx = {
+          TransactionType: 'AccountSet',
+          Account: account.address
+        }
+        if (flags[flag]) {
+          tx.SetFlag = ACCOUNT_FLAGS[flag]
+        } else {
+          tx.ClearFlag = ACCOUNT_FLAGS[flag]
+        }
+        transactions.push(tx)
```
ihomp, line 106: all the flags can be set in one transaction, you don't need to make a n arrow of tranactions

ihomp (changes requested):
some changes needed

pages/account-settings.js
ihomp: "You can enable and disable a combination of settings in one transaction using multiple tf flags, but not all settings have tf flags."

ihomp:
```
tfRequireDestTag
tfOptionalDestTag 
tfRequireAuth 
tfOptionalAuth 
tfDisallowXRP
tfAllowXRP
```

Those are all tf flags, they all can be set at ones

ihomp:
@pandablue0809 

It's ok to change flags in separate transactions; we just don't need to have an array of them to sign. 
As that is hard to test with hardware wallets, and mosyt likely some of them wouldn't support that. 

if you want sign each flag separately, you need to change UI. 

Instead of one button after all flags, you need to have separate buttons for each of the flags.

ihomp (commented):
- Add support for Xahau OR hide the page for Xahau users (for the first PR)
- Remove the transactions array; we can not submit an array of transactions.

pages/account-settings.js
pandablue0809: Hi @ihomp 
what one flag we use for xahau

pandablue0809:
hi @ihomp 
I fixed
plz give me feedback

pages/account-settings.js
ihomp: That is the list of flags that we need for xahau 

asfAccountTxnID
asfDefaultRipple
asfDepositAuth
asfDisableMaster
asfDisallowIncomingCheck
asfDisallowIncomingPayChan
asfDisallowIncomingTrustline
asfDisallowIncomingRemit
asfDisallowXRP
asfGlobalFreeze
asfNoFreeze
asfRequireAuth
asfRequireDest
asfTshCollect

ihomp:
Do you want to prepare only the flags on the fist PR?

There are many other settings, not only flags

2. fields (except flags)
https://xrpl.org/docs/references/protocol/transactions/types/accountset
https://docs.xahau.network/technical/protocol-reference/transactions/transaction-types/accountset

```
Domain
EmailHash
MessageKey
NFTokenMinter
TransferRate
TickSize
WalletLocator
```

Same fields for XAHAU, except `NFTokenMinter`

3. Account settings
https://docs.xahau.network/features/network-features/account-managment
different transaction types
`SetRegularKey`
`SignersListSet`

for Xahau optional `Import`

pages/account-settings.js
```diff
@@ -0,0 +1,355 @@
...
+  disallowIncomingCheck: 13,
+  disallowIncomingNFTokenOffer: 12,
+  disallowIncomingPayChan: 14,
+  disallowIncomingTrustline: 15,
+  xahauSpecificFlag: 99 // <-- placeholder value, replace with real one if known
+}
+
+export default function AccountSettings({ account, setSignRequest }) {
+  const { t } = useTranslation(['account-settings', 'common'])
+  const { theme } = useTheme()
```
ihomp, line 36: you don't need to import theme here

just specify the color in the css 
```

    my-element-name {
      color: "black";

      .dark & {
        color: #fff;
      }
    }
```

ihomp (changes requested):
need fixes

pandablue0809:
hi @ihomp 
how about implement only flags in this PR?

pandablue0809:
I fixed translation and theme

pages/account-settings.js
ihomp: I've described all the flags in the ticket

pages/account-settings.js
ihomp: I've specified all the flags in the ticket

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+
+// XRPL AccountSet flag values mapping
+const ACCOUNT_FLAGS = {
+  requireDestTag: 1,
+  disallowXRP: 3,
+  disallowIncomingCheck: 13,
+  disallowIncomingNFTokenOffer: 12,
+  disallowIncomingPayChan: 14,
+  disallowIncomingTrustline: 15,
+  xahauSpecificFlag: 99 // <-- placeholder value, replace with real one if known
```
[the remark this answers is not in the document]
ihomp: there are 3 flags for both networks that can be turned on/off in one transactions.

there are additional 9 flags that are common for both networks 

there are 3 more specific for XRPL
there are 2 more specific for XAHAU

Here you use specified values that can be enabled in transactions only one by one.

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+}
+
+export default function AccountSettings({ account, setSignRequest }) {
+  const { t } = useTranslation(['account-settings', 'common'])
+  const [loading, setLoading] = useState(true)
+  const [errorMessage, setErrorMessage] = useState('')
+  const [successMessage, setSuccessMessage] = useState('')
+  const [accountData, setAccountData] = useState(null)
+  const [flags, setFlags] = useState({
```
ihomp, line 39: the values will be different for each address, we can find the current ledger values through api with `ledgerInfo=true`
it's better to use null, rather then false as default - before you filled it in.

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+    disallowIncomingNFTokenOffer: false,
+    disallowIncomingPayChan: false,
+    disallowIncomingTrustline: false,
+    xahauSpecificFlag: false
+  })
+  const [selectedFlag, setSelectedFlag] = useState(null)
+  const [editedFlag, setEditedFlag] = useState(null)
+
+  // Determine which flags to use based on network
+  const flagKeys = xahauNetwork
```
ihomp, line 52: There are common and uncommon flags, you can check the ticket. 
common flags can be one array, which will be extended by network specific flags

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+      name: 'Disallow Incoming Payment Channels',
+      displayName: 'Allow Incoming PayChan',
+      status: (value) => (value ? 'Blocked' : 'Allowed')
+    },
+    disallowIncomingTrustline: {
+      name: 'Disallow Incoming Trust Lines',
+      displayName: 'Allow Incoming Trustline',
+      status: (value) => (value ? 'Blocked' : 'Allowed')
+    },
+    xahauSpecificFlag: {
```
ihomp, line 102: that need to be replaced, there is no such flag

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+        }
+        setLoading(false)
+      } catch (error) {
+        setErrorMessage(t('error.fetch'))
+        setLoading(false)
+      }
+    }
+    fetchAccountData()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [account, xahauNetwork])
```
ihomp, line 134: we don't need the `xahauNetwork` here
if it will be changed, it will redirect to a different website

pages/account-settings.js
```diff
@@ -0,0 +1,353 @@
...
+      </>
+    )
+  }
+
+  return (
+    <>
+      <SEO title={t('title')} />
+      <div className="content-center account-settings">
+        <h1 className="center">Account Settings</h1>
+        <p className="center">Manage your account settings on the XRP Ledger.</p>
```
ihomp, line 318: instead of "XRP Ledger" use` {explorerName}`  so it will accurate for xahau as well.

ihomp (changes requested):
need updates

pandablue0809:
hi @ihomp 
I fixed
plz check

ihomp:
it should be under **/services/** as well

pandablue0809:
hi @ihomp 
I fixed

pages/account-settings.js
```diff
@@ -0,0 +1,551 @@
...
+              </div>
+              <IoChevronForward size={20} className="icon-color" />
+            </div>
+          ))}
+        </div>
+        {errorMessage && <p className="red center">{errorMessage}</p>}
+        {successMessage && <p className="green center">{successMessage}</p>}
+        <br />
+        <div className="center">
+          <Link href={`/account/${account.address}`}>Back to my account</Link>
```
ihomp, line 546: show it only if there is `account.address` 
in cases when on mobile we redirected back from a wallet like shaman, it will be a new page, and not for user user will be loggedin

ihomp (changes requested):
small things

pages/account-settings.js
```diff
@@ -541,10 +542,14 @@ export default function AccountSettings({ account, setSignRequest }) {
         </div>
         {errorMessage && <p className="red center">{errorMessage}</p>}
         {successMessage && <p className="green center">{successMessage}</p>}
-        <br />
-        <div className="center">
-          <Link href={`/account/${account.address}`}>Back to my account</Link>
-        </div>
+        {account?.address && (
```
ihomp, line 545: no need to check it again, as you already add the check on line 397

ihomp (changes requested):
We need the UI that sgows all the buttons right away.

pandablue0809:
> @pandablue0809 it should also be under **/services/** like the rest tx pages

hi @ihomp 
I fixed
plz review again 🙏

ihomp:
@pandablue0809 we shouldn't have "optional" or "make optional" 
Required/Not required - Require/Don't require

ihomp:
The original screenshot in this thread had an explanation for each flag. 
we need to return it, we need some texts explaining what flags actually do.

ihomp:
**Authorized NFToken Minter** 
probably doesn't work without specifying the `NFTokenMinter` 
https://xrpl.org/docs/references/protocol/transactions/types/accountset#nftokenminter

so we need an **address** field to add `NFTokenMinter` , when we want to authrorise the minter.

ihomp:
if account has `ownerReserve > 0` (no trust lines, offers, escrows, payment channels, checks, or signer lists) etc - we need to disable the button to enable `TrustLine Clawback`. 

and we need to show why the button is disabled, same as for `Authorisation`. 

Allow account to claw back tokens it has issued. Can only be set if the account has an empty owner directory (no trust lines, offers, escrows, payment channels, checks, or signer lists). After you set this flag, it cannot be reverted. The account permanently gains the ability to claw back issued assets on trust lines.

ihomp:
May we can chnage the UI a bit

Like 

Flag name - status -                    [button to change]
Flag description. 

Destination Tag - Required                                                                        [Don't require]
If enabled, incoming transactions must include a Destination Tag.

If status is not default status, it should be highlighted by color, for example `className="orange" `

ihomp:
You can only enable the `No Freeze`, Once enabled, we should have button to disable it, as it is impossible. 

Also, if  `No Freeze` is enabled, we can not chnage "Global Freeze" - it should be disabled and should be explained why.

ihomp:
It seems **Deposit Authorization** status is shown differently on Account page on Account-settings page

ihomp:
Default Ripple status - not mathcing

PR #290, opened 2025-05-21
Update texts on the home's page hero section

pavel-fokin:
The goal is to improve SEO ranking by adding additional text that matches the keywords "xrp explorer" and "xrp scan".

**Desktop**
image


pages/index.js
```diff
@@ -39,7 +39,7 @@ const ldJsonWebsite = {
     nativeCurrency + ' Explorer',
     explorerName + ' Explorer',
     'Scan ' + nativeCurrency + ' Ledger',
-    'Bithomp' + explorerName + ' and Services'
+    'Bithomp ' + explorerName + ' and Services'
```
ihomp, line 42: don't we need a space after Bithomp? wouldn't it be "BithompXRP Ledger" without the space?

public/locales/en/common.json
```diff
@@ -572,6 +572,7 @@
   "explorer": {
     "header": {
       "main": "{{explorerName}} Explorer",
+      "sub": "Scan {{nativeCurrency}} transactions, accounts, tokens, and NFTs — the fast and trusted way to explore the {{nativeCurrency}} Ledger.",
```
ihomp, line 575: We usually keep the common.js files  and other files in sync with other currencies. so they all will have the same amount of lines. 

So, we have a few options here:
1) don't add translations yet, do not add it in this file and just keep it as text on on the page 
2) add it to all language files, translated by ai. 
3) if you want to prepare a namespace for future use, you can use it on the page like that with the fallback 
`{t('explorer.header.sub', { nativeCurrency }) || `Scan ${nativeCurrency} transactions, accounts, tokens, and NFTs — the fast and trusted way to explore the ${nativeCurrency}  Ledger`}`

styles/pages/index.scss
```diff
@@ -1,6 +1,8 @@
-h1.landing-h1,
-h2.landing-h2 {
-  height: 32.5px;
```
ihomp, line 3: I don't remember exactly, but I think the sizes were specified to make it better for SEO, so the things do not move on render - when font loads - the blocks were moved lower down. with the specified size the layout didn't move and it was better for SEO. 

but that was added before I removed the font we used, so now maybe it doesn't matter, but it's worth validating

ihomp (commented):
Some comments

styles/pages/index.scss
pavel-fokin: I'll try to validate. Font rendering issues can usually be fixed with something like `preload` or `font-display: swap`, if they will come up. Setting `height` feels a bit like workaround. Happy to leave as is if there's still a strong reason.

ihomp (approved):
good

PR #293, opened 2025-05-22
Load 'i18n-iso-countries' dynamically

pavel-fokin:
Reducing `app.js` bundle size by loading 'i18n-iso-countries' dynamically. Bundle size is reduced on ~148kB.

It affects pages:
`pages/nodes.js`
`pages/username.js`
`pages/validators.js`
`pages/admin/subscriptions.js`
`pages/nft/[[..id]].js`

### DIFF
--- a/components/Admin/BillingCountry.js
+++ b/components/Admin/BillingCountry.js
@@ -92,7 +92,7 @@ export default function BillingCountry({ billingCountry, setBillingCountry, choo
           {billingCountry && (
             <>
               Your billing country is{' '}
-              <a onClick={() => setChoosingCountry(true)}>{countries.getNameTranslated(billingCountry)}</a>
+              <a onClick={() => setChoosingCountry(true)}>{countries?.getNameTranslated?.(billingCountry) || billingCountry}</a>
             </>
           )}
         </>
--- a/components/Admin/BillingCountry.js
+++ b/components/Admin/BillingCountry.js
@@ -10,7 +10,15 @@ import { axiosAdmin } from '../../utils/axios'
 export default function BillingCountry({ billingCountry, setBillingCountry, choosingCountry, setChoosingCountry }) {
   const router = useRouter()
   const { i18n } = useTranslation()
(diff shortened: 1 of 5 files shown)


ihomp:
That's needed! I was looking at the bundle another day, and I was thinking that not all translations there are needed from the first load for sure =) 

Let me know if it's ready for review!

ihomp:
When I open a page `username` I see now a country code instead of a translated country name. 
Right now, we have a translated country name:
http://bithomp.com/en/username 

Screenshot 2025-05-23 at 15 47 20

And now:
Screenshot 2025-05-23 at 15 47 12

The countries list is loaded and available in the dropdown; we just need to update the value I guess.

ihomp:
I'll prepare an update

pavel-fokin:
I cannot reproduce the bug you've mentioned btw

ihomp:
I have also another bug that country is not saved in local storage.

pavel-fokin:
> I have also another bug that country is not saved in local storage.

On what page?

ihomp:
> > I have also another bug that country is not saved in local storage.
> 
> On what page?

I've tested only the `/username` page

ihomp:
With my commit, it works for me:
- to see a translated country name inside of country select 
- to save the country in localstorage 
- to show the country saved in localstorage 
- if no country in localStorage: set and show the country by IP address 

But so far I only tested the `username` page only.

ihomp:
On the page `/nodes` the UI is now changing from country codes to country names 
- I guess, we can leave as it is, though the jumps probably not so good for the seo page. 

Also, right now the loaded countries are not used between the page, ones loaded on one page, they will be loaded again on another page where it's needed - we could reuse the loaded data - but it's not a priority right now, so we can leave it as it is.

ihomp:
tested:
- usernames 
- validators
- nodes 
- nft/ nft with evernodeRegistartion
- admin/subscriptions 

looks ok

PR #295, opened 2025-05-23
Update pricing section on the home page

pavel-fokin:
Add title and update layout for the pricing section.

### How it's now
image

--- 

### How it's suggested
image

---
---

### How it's now
image

---

### How it's suggested
image

### DIFF
--- a/styles/pages/index.scss
+++ b/styles/pages/index.scss
@@ -1,7 +1,19 @@
-.home-hero {
+.home-section {
   padding: 20px 0;
-  p {
+  > p {
     margin: 20px;
+  }
+}
+
+.home-price-tools {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  align-items: center;
+  gap: 10px;
+
+  @media only screen and (max-width: 768px) {
+    flex-direction: column;
   }
 }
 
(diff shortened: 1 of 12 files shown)


styles/pages/index.scss
```diff
@@ -1,17 +1,29 @@
...
+}
+
 .home-converter {
   margin-left: 20px;
   z-index: 2; //to show over the chart
   padding-top: 32px;
   height: 200px;
   text-align: center;
-  @media only screen and (max-width: 803px) {
+  @media only screen and (max-width: 768px) {
```
pavel-fokin, line 26: I changed it as `768px` is a more standard breakpoint. Can change back.

pavel-fokin: what was the width you've tested with?

pavel-fokin: nevermind the question above :) i'll try to fix it

pavel-fokin: I fixed the issue for widths between 768–803px. I'm trying to stick to standards and avoid workarounds unless absolutely necessary, as they can make the code harder to maintain. Let me know what you think!

image

ihomp: I like the standards, but then we need to make updates in the way so nothing gets worse than it used to be. It takes time to test and change the code in several places. Custom widths can be more beneficial for UI, as they are specific for our UI elements, and they will work on all different kinds of screens... but it takes a longer time to test.

PR #296, opened 2025-05-23
some new lines are translated

Anna15170221:
partially translated - will create a separate PR for remaining lines.

### DIFF
--- a/pages/faucet.js
+++ b/pages/faucet.js
@@ -46,7 +46,7 @@ export async function getServerSideProps(context) {
 }
 
 export default function FaucetPage({ account, showAds, sessionTokenData }) {
-  const { t } = useTranslation(['faucet', 'common'])
+  const { t } = useTranslation()
 
   return (
     <>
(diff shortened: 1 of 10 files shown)


pages/faucet.js
```diff
@@ -46,7 +46,7 @@ export async function getServerSideProps(context) {
 }
 
 export default function FaucetPage({ account, showAds, sessionTokenData }) {
-  const { t } = useTranslation()
+  const { t } = useTranslation(['faucet', 'common'])
```
ihomp, line 49: doesn't it work without it? it should specified in ServerSide a bit more up

pages/faucet.js
```diff
@@ -67,7 +67,7 @@ export default function FaucetPage({ account, showAds, sessionTokenData }) {
       />
       <div className="content-text content-center">
         <h1 className="center">
-          {explorerName} {t('menu.developers.faucet')} — Get Free {nativeCurrency}
+          {explorerName} {t('common:menu.developers.faucet')} — {t('faucet:get-free')} {nativeCurrency}
```
ihomp, line 70: for common, you probably don't need to add "common:", for faucet, we usually add `{t('get-free', {ns: 'faucet'})} ` I like the format {t('faucet:get-free')}, may be we can start usuing it

ihomp (commented):
some comments

Issue #282, opened 2025-05-14
New Page - Send Payments


Anna15170221:
Create a new page where the user can send payments in XRP (and XAH for Xahau).

ihomp:
The specs are here:
https://xrpl.org/docs/references/protocol/transactions/types/payment

we have the send XRP functionality in old explorer bithomp.com/explorer/ when user is logged in with a hardware wallet, so we need to replace it with a new and more advanced one. 

As an inspiration, you can look at xrptoolkit.com/send 

The fields we need (MINIMUM)
Destination - search address field 
DestinationTag - digits only (same as on the testPayment/faucet page) 
Amount - for the first PR, it can be XRP only. (IOU can be added later.) (example in faucet)
Message - Memo should be added as text. 

fields hidden under **advanced payment**  checkbox: 
Fee - user should be able to specify a fee, same field as for amount (example in faucet), Max Fee should be 1 XRP -otherwise, give an error. 

As the 2 step (separate PR)

We need to allow sending IOU (Tokens) as well. 

- We can check what the Tokens destination has
- We can check what Tokens the user has 
So if both Source and Destination have the same one, we can offer to send it. 

As a step 3 (separate PR): Allow conversion payments and give a choice of which asset to convert for the payment. (We have that functionality in the bithomp-tools repository.)

ihomp:
We also need two fields under advanced options:

`SourceTag`: // same format as DestinationTag
'InvoiceID'

PR #297, opened 2025-05-23
New Page - Send Payments

pandablue0809:
# Issue
[Create a new page where the user can send payments in XRP (and XAH for Xahau).](https://github.com/Bithomp/frontend-react/issues/282)
# Description
- [x] first step 

image

### DIFF
--- a/components/SignForm.js
+++ b/components/SignForm.js
@@ -511,7 +511,8 @@ export default function SignForm({
     }
   }
 
-  const checkTxInCrawler = async ({ txid, redirectName }) => {
+  const checkTxInCrawler = async ({ txid, redirectName, txType }) => {
+    //TODO: Check the case for failed transactions.
     setAwaiting(true)
     setStatus(t('signin.status.awaiting-crawler'))
     //txid can be in the ledger or not, so we need to check it in the ledger
@@ -549,9 +550,21 @@ export default function SignForm({
               closeSignInFormAndRefresh()
             }
             return
+          } else if (TransactionType === 'Payment') {
+            if (signRequest?.callback) {
+              signRequest.callback({
+                result: response.data
+              })
+            }
+            closeSignInFormAndRefresh()
+            return
           }
           checkCrawlerStatus({ inLedger: includedInLedger, type: TransactionType })
         } else {
+          if (txType === 'Payment') {
+            closeSignInFormAndRefresh()
+            return
+          }
           //if not validated or if no ledger info received, delay for 1.5 seconds
           delay(1500, checkTxInCrawler, { txid, redirectName })
         }
@@ -652,24 +665,16 @@ export default function SignForm({
     }
 
     // For NFT and DID transaction, lets wait for crawler to finish it's job
-    if (txType?.includes('NFToken') || txType?.includes('URIToken') || txType?.includes('DID')) {
-      checkTxInCrawler({ txid: txHash, redirectName })
+    if (
+      txType?.includes('NFToken') ||
+      txType?.includes('URIToken') ||
+      txType?.includes('DID') ||
+      txType === 'Payment'
+    ) {
+      checkTxInCrawler({ txid: txHash, redirectName, txType })
       return
     } else {
-      // For payment transactions, get the transaction result and pass it to the callback
-      if (txType === 'Payment' && signRequest?.callback) {
-        try {
-          const response = await axios('xrpl/transaction/' + txHash)
-          if (response.data) {
-            signRequest.callback({
-              result: response.data
-            })
-          }
-        } catch (error) {
-          console.log('Error fetching transaction result:', error)
-        }
-      }
-      // no checks or delays for non NFT/DID transactions
+      // no checks or delays for non NFT/DID/Payment transactions
       closeSignInFormAndRefresh()
     }
   }
--- a/pages/services/send.js
+++ b/pages/services/send.js
@@ -94,6 +94,7 @@ export default function Send({ account, setSignRequest }) {
         callback: (result) => {
           if (result.result) {
             setTxResult({
+              status: result.result.meta?.TransactionResult,
               date: result.result.date,
               destination: result.result.Destination,
               amount: amountFormat(result.result.Amount),
@@ -207,7 +208,7 @@ export default function Send({ account, setSignRequest }) {
               Send Payment
             </button>
           </div>
-          {txResult && (
+          {txResult?.status === 'tesSUCCESS' && (
             <>
               <br />
               <div>


pandablue0809:
hi @ihomp 
I completed firt step
plz review

pages/send.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+    try {
+      const payment = {
+        TransactionType: 'Payment',
+        Account: account.address,
+        Destination: address,
+        Amount: xrpToDrops(amount.toString()),
+        Memos: memo ? [{
+          Memo: {
+            MemoData: Buffer.from(memo).toString('hex').toUpperCase(),
```
ihomp, line 79: I think we have a function `encode` in utils for this.

pages/send.js
```diff
@@ -0,0 +1,252 @@
...
+            MemoFormat: Buffer.from('text/plain').toString('hex').toUpperCase()
+          }
+        }] : undefined,
+        DestinationTag: destinationTag ? parseInt(destinationTag) : undefined,
```
ihomp, line 83: destinationTag can be 0 and it would be valid 

It's better to add some fields conditionally

```
if (destinationTag || destinationTag==='0') {
   payment.DestinationTag = destinationTag
}
```

the same  kind of conditions for the memo and the fee

pages/send.js
```diff
@@ -0,0 +1,252 @@
...
+import CheckBox from '../components/UI/CheckBox'
+import AddressInput from '../components/UI/AddressInput'
+import FormInput from '../components/UI/FormInput'
+import CopyButton from '../components/UI/CopyButton'
+import { LinkTx, LinkAccount } from '../utils/links'
+import NetworkTabs from '../components/Tabs/NetworkTabs'
+import { typeNumberOnly, isAddressValid, isTagValid , nativeCurrency} from '../utils'
+import { fullDateAndTime , timeFromNow } from '../utils/format'
+import { useState } from 'react'
+import { xrpToDrops } from 'xrpl'
```
ihomp, line 14: I don't think we use this library pls do not add it.

pages/send.js
```diff
@@ -0,0 +1,252 @@
...
+      setFeeError(t('form.error.max-fee', 'Maximum fee is 1 ' + nativeCurrency))
+    } else {
+      setFeeError('')
+    }
+  }
+
+  const handleSend = async () => {
+    setError('')
+    setTxResult(null)
+    console.log('Starting transaction...')
```
ihomp, line 45: clean up the logs

pages/send.js
```diff
@@ -0,0 +1,252 @@
...
+
+    // Validate required fields
+    if (!address || !isAddressValid(address)) {
+      console.error('Invalid address:', address)
+      setError(t('form.error.invalid-address', 'Please enter a valid destination address'))
+      return
+    }
+
+    if (!amount || isNaN(parseFloat(amount)) || parseFloat(amount) <= 0) {
+      console.error('Invalid amount:', amount)
```
ihomp, line 55: Clean up the console.

pages/send.js
pandablue0809: you mean xrpToDrops ?

ihomp: I mean the library 'xrpl', I don't see we have it installed, and I don't see you installed it in that PR. But anyway, we don't need it.

ihomp:
Please don't add translation placeholders, unless you add them in translation files. 

`{t('transaction.sequence', 'Sequence')}`

should be just `Sequence`

pandablue0809:
I fixed @ihomp 
plz check 🙏

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+        payment.Memos = [{
+          Memo: {
+            MemoData: encode(memo),
+            MemoFormat: encode('text/plain')
+          }
+        }]
+      }
+
+      if (fee) {
+        payment.Fee = String(Math.round(parseFloat(fee) * 1000000))
```
ihomp, line 86: User shouldn't be able to enter more than 6 digits after the comma. 
then we don't need Math.round

If you want to use precise calculations, we have `multiply` from `utils/calc.js`

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+
+      setSignRequest({  
+        request: payment,
+        wallet: account.wallet,
+        callback: (result) => {
+          if (result.result) {
+            setTxResult({
+              date: result.result.date,
+              destination: result.result.Destination,
+              amount: (parseInt(result.result.Amount) / 1000000).toString(),
```
ihomp, line 97: We have functions called `amountFormat` from `utils/format.js` that converts amounts

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+              destinationTag: result.result.DestinationTag?.toString(),
```
ihomp, line 98: why do you need to make them strings?

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+              sourceTag: result.result.SourceTag?.toString(),
```
ihomp, line 99: why do you need to make them strings?

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+              fee: (parseInt(result.result.Fee) / 1000000).toString(),
+              sequence: result.result.Sequence?.toString(),
```
ihomp, line 101: why do you need to make them strings?

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+      })
+    } catch (err) {
+      setError(err.message)
+    }
+  }
+
+  return (
+    <>
+      <SEO
+        title={'Send Payment'}
```
ihomp, line 124: you don't need `{}` for simple strings

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+        description={'Send a payment to a destination address'}
```
ihomp, line 125: you don't need `{}` for simple strings

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+      />
+      <div className="content-text content-center">
+        <h1 className="center">Send Payment</h1>
+        <NetworkTabs />
+        
+        <div>
+          <AddressInput
+            title={t('table.destination', 'Destination')}
```
ihomp, line 133: no need to specify fallback, as we have it in all languages translated

pages/send.js
```diff
@@ -0,0 +1,262 @@
...
+              min="0"
+              type="text"
+              inputMode="decimal"
+              defaultValue={amount}
+            />
+          </div>
+          <div className="form-input">
+            {width > 1100 && <br />}
+            <span className="input-title">
+              {t('table.memo', 'Memo')}
```
ihomp, line 170: no need for fallback

ihomp:
1) All the field values should be added in the URI, and they also need to be read from the URI, the same way as you did for `uri` and `taxon` on the NFT MINT page 
- that can be added in a separate PR

2) we need links in the menu services - can be done in a separate PR

ihomp:
- send should be inside the services 
`/services/send`

ihomp (approved):
The code looks great

ihomp:
@Anna15170221 can you please test it?

ihomp:
@Anna15170221 
- We need to test the flow on the mobile with the Xaman wallet 
- Test with BiFrost or GirinWallet mobile/desktop 
- Test on Xahau network 

I haven't tested on mobile.. it seems we won't even show the transaction hash there. 
It would be good to show the link, like we do after the nft mint.

ihomp:
I'll merge this, but 
1) we will need update for mobile, because after xaman signing, there is no `signRequest` available. 
2) It won't be public, as we don't have links in the menus in this PR.

PR #304, opened 2025-05-27
we need link of send payment page in the menu services

pandablue0809:
# Issue
[we need links in the menu services](https://github.com/Bithomp/frontend-react/issues/282)


pandablue0809:
@ihomp 
plz review

PR #305, opened 2025-05-28
Add tailwindcss

pavel-fokin:
As we plan to add more textual content such as articles and posts, this is a good opportunity to start using `tailwindcss`, specifically the `@tailwindcss/typography` plugin, on the existing **learn pages**.

> ⚠️ DISCLAIMER ⚠️
> We’re not planning to use Tailwind classes unsystematically across the entire project, as that could lead to unmaintainable styling. Instead, we aim to gradually introduce Tailwind in selected components and pages, leveraging its sensible defaults and scalable design system.

## How it was
image

## How it's suggested
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "@paypal/react-paypal-js": "^8.1.3",
     "@sentry/nextjs": "^9.22.0",
     "@svgr/webpack": "^8.1.0",
+    "@tailwindcss/postcss": "^4.1.7",
     "@trezor/connect-web": "^9.4.3",
     "@walletconnect/modal-sign-react": "^2.7.0",
     "@walletconnect/utils": "^2.20.3",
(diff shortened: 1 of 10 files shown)


pavel-fokin:
> am I missing something important?

You points are valid. The goal is to adopt tailwind in areas where it has clear advantages:
- textual content rendering with @tailwindcss/typography, like for blog posts and learn sections
- faster layout prototyping for new pages
- better design consistency with utility classes for spacing, layout, and responsive breakpoints

Also I'd say that tailwind has a very low learning curve, is widely adopted, and tools like copilot help a lot. We’re starting small and scoped and can evaluate together if it proves useful or not.

Let me know what you think.

Anna15170221:
@ihomp, I can't test locally for some reasons, need your help )

Anna15170221:
@pavel-fokin 
@ihomp 
the blog pages look fine, but all the images are now of different width - not sure if it's an issue.

PR #306, opened 2025-05-28
All the field values should be added in the URI

pandablue0809:
# Issue
All the field values should be added in the URI, and they also need to be read from the URI, the same way as you did for uri and taxon on the NFT MINT page


pandablue0809:
hi @ihomp 
plz review 🙏

pages/services/send.js
```diff
@@ -133,6 +181,8 @@ export default function Send({ account, setSignRequest }) {
             name="destination"
             hideButton={true}
             setValue={setAddress}
+            rawData={isAddressValid(address) ? { address } : {}}
```
ihomp, line 184: does it do anything? Why do we need it here?

pandablue0809: I think when the address is reflected in the URL, it automatically fills in the field.

pages/services/send.js
```diff
@@ -12,21 +12,69 @@ import { multiply } from '../../utils/calc'
...
   const width = useWidth()
-  const [address, setAddress] = useState('')
-  const [destinationTag, setDestinationTag] = useState('')
-  const [amount, setAmount] = useState('')
-  const [memo, setMemo] = useState('')
-  const [showAdvanced, setShowAdvanced] = useState(false)
-  const [fee, setFee] = useState('')
+  const router = useRouter()
+  const [address, setAddress] = useState(isAddressValid(addressQuery) ? addressQuery : null)
+  const [destinationTag, setDestinationTag] = useState(isTagValid(destinationTagQuery) ? destinationTag : null)
```
ihomp, line 31: shouldn't be destinationTag as destinationTagQuery? 

`isTagValid(destinationTagQuery) ? destinationTagQuery`

pages/services/send.js
```diff
@@ -175,7 +225,7 @@ export default function Send({ account, setSignRequest }) {
               defaultValue={memo}
             />
           </div>
-          <CheckBox checked={showAdvanced} setChecked={() => setShowAdvanced(!showAdvanced)} name="advanced-payment">
+          <CheckBox checked={showAdvanced} setChecked={() => {setShowAdvanced(!showAdvanced), setFee(null)}} name="advanced-payment">
```
ihomp, line 228: why do you do setFee(null) here?

pages/services/send.js
ihomp: I've fixed it

ihomp (approved):
Do not allow submit form if the fee is over 1 (specified in uri)

ihomp:
@Anna15170221 can you plese test this one? 

- make sure adding and removing fee is working 
- make sure that all added values are added in the uri
- if you chnage values in the uri, all the fields will be updated (if they are correct)

pages/services/send.js
pandablue0809: If you enter `Fee` and set `setShowAdvanced` to `false`, the `Fee` will be included in the transaction.

PR #307, opened 2025-06-02
Attempt to fix issue with favicon in production

pavel-fokin:


ihomp:
Would that only fix for favicon? We need all files in the public folder to be accessible in the same way.

ihomp:
it seems other works already, right?
http://localhost:3000/images/checkmark.svg

pavel-fokin:
It’s hard to reproduce this issue locally and it’s works occasionally in
prod. So just want to check if NextResponse.next() helps. Otherwise will
need to research more.

Issue #313, opened 2025-06-05
Create a new /amendment/ page to display validator votes


Anna15170221:
Create a page at /amendment/{amendment_name} to show which validators are voting for and which are not voting or voting against a specific amendment. (Yeas/ Nays)

Current implementation example:
https://bithomp.com/en/validators?amendment=DynamicNFT
All necessary data is already available on current pages — it just needs to be reorganized and presented on separate pages, for example: 
https://bithomp.com/en/amendment/DynamicNFT

ihomp:
<img width="450" height="314" alt="Image" src="https://github.com/user-attachments/assets/d0fc3ff2-f187-4109-a8ed-b1540f50fee6" />

we should fix css for small icons like flags etc, so they lines with them won't take more space then normal lines wthout icons...

the first looks compact, the ones with icons should as compact as the first one in the list, no extra space beeween lines

ihomp:
Estimated activation time 

we don't need to show fields in the table that are undefined, because there are no such for already activated or obsolete

ihomp:
[...]

PR #316, opened 2025-06-06
Create a new /amendment/ page to display validator votes

pandablue0809:
# Issue
[Create a page at /amendment/{amendment_name} to show which validators are voting for and which are not voting or voting against a specific amendment. (Yeas/ Nays)](https://github.com/Bithomp/frontend-react/issues/313)

### DIFF
--- a/pages/amendment/[amendmentName].js
+++ b/pages/amendment/[amendmentName].js
@@ -127,9 +127,6 @@ export default function AmendmentSummary({ amendmentName, amendmentData, feature
   const showHash = (hash) => {
     return windowWidth > 1140 ? <>{hash} </> : <>{shortHash(hash)} </> 
   }
-  console.log('nays', nays)
-  console.log('featureData', featureData)
-  console.log('amendmentData', amendmentData)
   const fixCountry = (country) => {
     //accept UK as a country code for GB
     return country?.toUpperCase() === 'UK' ? 'GB' : country


Anna15170221:
no, the same for me.

pandablue0809:
hi @ihomp 
plz review again

ihomp:
- Yes / Nays, on desktop, should be two columns next to each other 
- Amendments with majority should show status that they are reached majourity 
- not enabled amendments like `fixInvalidTxFlags` should also show threshold 
- Yes/nays counts and list are incorrect for `fixInvalidTxFlags`, should be 6/29 not 0/35 
- In the list of validators for Yes /Nays  we need to have country flag, name, domain - similiar to what we have in /validators page
- on the testnet of not all data is available UI should look good too.

ihomp:
- Yes / Nays, on desktop, should be two columns next to each other 
- Amendments with majority should show status that they are reached majourity 
- not enabled amendments like `fixInvalidTxFlags` should also show threshold 
- Yes/nays counts and list are incorrect for `fixInvalidTxFlags`, should be 6/29 not 0/35 
- In the list of validators for Yes / Nays  we need to have country flag, name, domain - similiar to what we have in /validators page
- on the testnet, if not all data is available, the UI should look good too.

ihomp:
For already enabled amendments, we do not have hitorical data who voted for it.. 
so no need to show yes/nays 
en/amendment/AMMClawback example

ihomp:
inspiration for data https://livenet.xrpl.org/amendments

ihomp (approved):
It's a bit raw/extensive, but I'll merge it. Fixes can be done separately.

PR #328, opened 2025-06-11
Fix style of tokenSelector

pandablue0809:
# Description
- we need z-index higher, as nothing should be on top of it. I’ve tested on page amm, and the line was over the popup
- for the icons hashicons in the circles doesn’t look so good, we need to fix css to enlarge hashicons, so they will look like in circle shapes too, like all the rest.
- when searching for new coins, the size of the popup chnages.. we need to keep the same size of popup it shouldn’t jump

we should have a message saying that there are more results, and the user need to specify an issuer to narrow down the search -


ihomp:
Screenshot 2025-06-13 at 12 13 40
I still see the line on top of it

ihomp:
On initial open, when I scroll down all the currencies.. I don't see the message to specify the search.. as there are more than 100

ihomp (approved):

PR #329, opened 2025-06-11
Add token selector in amms page

pandablue0809:
# Description
add tokenSelector on the page /amms instead of xrp/rlusd currrency selector

### DIFF
--- a/pages/amms.js
+++ b/pages/amms.js
@@ -129,6 +129,9 @@ export default function Amms({
     return ''
   })
 
+  // control radio selection: 'all' | 'single'
+  const [filterMode, setFilterMode] = useState(() => (token?.currency ? 'single' : 'all'))
+
   const controller = new AbortController()
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


styles/ui.scss
```diff
@@ -87,7 +87,7 @@ body.dark {
       border-top: 1px solid #ccc;
 
       @media(min-width: 1300px) {
-        z-index: 4;
```
ihomp, line 90: what this for? looks dangerous

styles/components/tokenSelector.scss
```diff
@@ -197,21 +195,19 @@
...
     justify-content: center;
     overflow: hidden;
   }
 
-  &-modal-icon {
-    width: 28px;
-    height: 28px;
-    object-fit: cover;
-  }
+  // &-modal-icon {
```
ihomp, line 208: commented staff should be better removed

pandablue0809: done

styles/ui.scss
ihomp: ok, I've checked it seems effect only the line.. we can keept at 3.

ihomp (approved):
- Token selector send requests to tokens, even though it wasn't clicked yet. 
- The token selector doesn't look like a popup on the dark mode. 
- The fixes can be done in a seprate PR

pages/amms.js
```diff
@@ -142,29 +114,41 @@ export default function Amms({
...
+        addAndRemoveQueryParams(router, [{ name: 'currency', value: nativeCurrency }], ['currencyIssuer', 'service', 'username'])
+      } else if (token.issuer) {
+        const params = [
+          { name: 'currency', value: token.currency },
+          { name: 'currencyIssuer', value: token.issuer }
+        ]
+        addQueryParams(router, params)
       } else {
-        removeQueryParams(router, ['currencyIssuer', 'currency'])
+        removeQueryParams(router, ['currencyIssuer', 'currency', 'service', 'username'])
```
ihomp, line 145: , 'service', 'username' - can not be in params

ihomp:
Screenshot 2025-06-17 at 04 45 25

Rather, we need to have two options:
- All tokens
- Single token

And if all tokens are chosen, hide the token selector. 

Option two is to integrate "All tokens" or "Non-selected" inside the token selector as a default option.

so it will be like:
Token: 
"Non-selected" - means "all" - no filter 
or a choosen Token.

ihomp (changes requested):
we need small UI fixes

ihomp (approved):
When selecting SingleToken, the token is already selected as xrp - but the view is not updated

PR #330, opened 2025-06-11
Article how to mint nft on xrpl and xahau

Anna15170221:


ihomp (changes requested):
1) "Minting NFTs typically requires XRP held in reserve — unless you mint and create the sell offer in the same transaction," - even if you create a sell offer - it still requires a reserve 
2) "Paste your IPFS link." - you can specify that this shhould be the link to your metadata  file
3) transferable, burnable, and whether it will be sold for XRP only. - for xahau - it can be only burnable - can be described too. 
4) transferable, burnable, and whether it will be sold for XRP only. - you can add Mutable  (change URI) for xrpl only. 
5) it would be nice to add a link to /account-settings to add the minter - describe 
6) or mint on behalf of another account. - it is not possible on xahau - that should be removed 
7) "You can also update the URI of your Dynamic NFTs." - that is not possible on xahau - should be removed. 
8) on page NFT mint we need to add link to the article 
9) later we need to include this page to the learn section.

ihomp (approved):

PR #334, opened 2025-06-12
Fix/UI account setting page

pandablue0809:


ihomp:
My account has enabled rippling, but the page shows rippling is disabled. 
"Disabled" is shown in orange, though "disabled" is the default state, and it shouldn't be orange. 

The warning for master key we can copy from xrpl.org

> Warning
> Disabling the master key pair removes one method of authorizing transactions. You should be sure you can use one of the other ways of authorizing transactions, such as with a regular key or by multi-signing, before you disable the master key pair. (For example, if you assigned a regular key pair, make sure that you can successfully submit transactions with that regular key.) Due to the decentralized nature of the XRP Ledger, no one can restore access to your account if you cannot use the remaining ways of authorizing transactions.

Also we need to add notes: 

> You should do this if your account's master key pair may have been compromised, or if you want to make multi-signing the only way to submit transactions from your account.

> To disable the master key pair, you must use the master key pair. However, you can re-enable the master key pair using any other method of authorizing transactions.

pandablue0809:
@ihomp 
Plz review again 🙏

ihomp:
Sign In to your account, should not link to the old explorer website 
"Sign In to your account" should be the same as "Sign in" on the page `/username ` near the address account. 
Or like in the munu "SignIn" - its open the popup to signin.

ihomp:
Advanced options should be clickable when user is not signed in.

ihomp (changes requested):
We need some chnages as in comments.

pages/services/account-settings.js
```diff
@@ -244,28 +254,36 @@ export default function AccountSettings({ account, setSignRequest }) {
       try {
         const response = await axios(`/v2/address/${account.address}?ledgerInfo=true`)
         setAccountData(response.data)
-
+        console.log('response.data', response.data)
```
ihomp, line 257: we need clear logs

pages/services/account-settings.js
```diff
@@ -300,17 +319,18 @@ export default function AccountSettings({ account, setSignRequest }) {
   }
 
   const canEnableRequireAuth = () => {
-    // Can only be enabled if the address has no trust lines connected to it
-    // We can check this by looking at trustline count or ownerReserve for trust lines
-    return accountData?.ledgerInfo?.ownerReserve === 0 || !accountData?.ledgerInfo?.ownerReserve
+    // Can only be enabled if the account has no owner objects (trustlines, offers, escrows, etc.)
+    const ownerReserveZero = accountData?.ledgerInfo?.ownerReserve === 0 || !accountData?.ledgerInfo?.ownerReserve
```
ihomp, line 323: does the v2/address has `ownerReserve`? 
according to our docs
https://docs.bithomp.com/#get-information-for-an-address
we should have there `ownerCount`

pages/services/account-settings.js
```diff
@@ -486,24 +506,35 @@ export default function AccountSettings({ account, setSignRequest }) {
...
     let buttonDisabled = false
     let disabledReason = ''
-
+    
     // Check specific conditions for disabling buttons
     if (flag === 'allowTrustLineClawback' && !canEnableTrustLineClawback() && !currentValue) {
       buttonDisabled = true
       disabledReason =
-        'Can only be enabled if account has no trustlines, offers, escrows, payment channels, checks, or signer lists (ownerReserve must be 0)'
+        'Can only be enabled if account has no trustlines, offers, escrows, payment channels, checks, or signer lists (issuedTokens will 0, if ownerCount is 0)'
```
ihomp, line 529: Omit "(issuedTokens will 0, if ownerCount is 0)" 
remove it, users don't need to see it.

you already wrote "no trustlines, offers, escrows, payment channels, checks, or signer lists" - that means the same.

pages/services/account-settings.js
```diff
@@ -486,24 +506,35 @@ export default function AccountSettings({ account, setSignRequest }) {
...
     }
-
+    
     if (flag === 'requireAuth' && !canEnableRequireAuth() && !currentValue) {
       buttonDisabled = true
-      disabledReason = 'Can only be enabled if account has no trust lines connected to it'
+      disabledReason =
+        'Can only be enabled if the account has no trustlines, offers, escrows, payment channels, checks, or signer lists (issuedTokens will 0, if ownerCount is 0)'
```
ihomp, line 535: delete "(issuedTokens will 0, if ownerCount is 0)"

ihomp (changes requested):
do we really have ownerReserve instead of ownerCount? we need to use ownerCount. and if we have ownerReserve we need to update docs

Issue #225, opened 2025-03-11
Show more details and time machine for Object page


ildaruz:
I would like have an option to see more detailed information about ledger entry objects.
Use new `/v2/ledgerEntry/:index` to get object information

Possible types 
```
AccountRoot
Amendments
Check
DepositPreauth
DirectoryNode
Escrow
FeeSettings
LedgerHashes
NFTokenOffer
NFTokenPage
NegativeUNL
Offer
PayChannel
RippleState
SignerList
Ticket

AMM
PermissionedDomain
Credential
Oracle
DID
MPToken
MPTokenIssuance

URIToken
UNLReport
Hook
HookState
HookDefinition
ImportVLSequence

Bridge
XChainOwnedClaimID
XChainOwnedCreateAccountClaimID
```

Extended details is available on
```
RippleState
AccountRoot
Offer
Check
DirectoryNode
PayChannel
```

endpoint adds extra details about 
addresses(Account, Owner, Destination, SendMax, issuer) details
Flags parsing as `flags` object
Expiration as `expiration` - time in seconds
CancelAfter as `cancelAfter` - time in seconds

Allow to use chose previous version by date/time pickup send it as ISO time or in seconds `2014-06-19T13:57:10.000Z` `1403186230`
```
/v2/ledgerEntry/0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8?date=1403186230
```

Page should allow to go back to previous version of the object.
if `PreviousTxnLgrSeq` available it should allow with single action to get previous one by ledger 
```
/v2/ledgerEntry/0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8?ledgerIndex=1403186230
```

in case requested version and version in response has the same  ledger index - request previous one by `PreviousTxnID`
```
/v2/ledgerEntry/0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8?previousTxHash=5B37F9EE650016C27EF3DEC96159D1F4C4DB6E9AFFB6A0228A669B9AB4C0A30A
```

In case `date`, `ledgerIndex` or `previousTxHash` is used, the have to be saved as params for object page

Option to see raw objects, it require exclude fields added by the `/v2/ledgerEntry/:index`  or use `/xrpl/ledgerEntry/:index` to get one

ildaruz:
LockedBalance is not showing balance in correct way
[...]

PR #335, opened 2025-06-12
Show more details and time machine for Object page

pandablue0809:
# issue
[have an option to see more detailed information about ledger entry objects.](https://github.com/Bithomp/frontend-react/issues/225)


ihomp (changes requested):
I've tried to open a page, it errors 
`Error: Objects are not valid as a React child (found: object with keys {currency, issuer, value, issuerDetails}). If you meant to render a collection of children, use an array instead.`

so, it brakes. 

Error: Objects are not valid as a React child (found: object with keys {currency, issuer, value, issuerDetails}). If you meant to render a collection of children, use an array instead.

I've tried this 

/en/object/0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8

pandablue0809:
I fixed @ihomp 
plz review again 🙏

ihomp:
I guess when we get 

```
{
    "ledger_entry": "0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8",
    "error": "entryNotFound",
    "status": "error",
    "validated": true
}
```

we don't need to show the JSON, we can just show the error message that Such Object is not found on that network.

ihomp:
when I press previous by tx or by ledger- nothing is happening.  

en/object/0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8

ihomp:
we shouldn't do an api call to get the raw data initially; we just need to make a call to get the raw data when the user clicks "Show the raw Json data". Otherwise it's an unnecessary extra call.

pandablue0809:
hi @ihomp 
I fixed
plz review again

ihomp:
I like the idea of having icons for issued tokens infrount of the amounts..
But we can't use that until we have it secure/verifiable.. 
Right now anyone can set an avatar/logo and pretend to be any coin.

Let's remove the icons before the amounts, so people won't be tricked by fake icons.

ihomp (changes requested):
fixes are required

pandablue0809:
@ihomp 
I fixed
plz check 🙏

styles/ui.scss
```diff
@@ -877,4 +877,4 @@ body.dark {
 @include meta.load-css("pages/governance.scss");
 @include meta.load-css("pages/pro-history.scss");
 @include meta.load-css("pages/account-settings.scss");
-
+@include meta.load-css("pages/object.scss");
```
ihomp, line 880: We shouldn't add the page styles to the general bundle. 

we should use modules 
Check how it is done for `component/tiles `
there is css in `tiles.module.css `

That way css will be only loaded for that specific page without increasing the initial bundle. 

We should decrease the size of the bundle, not increase it. 

As alternative, the css can also be added on page right away, using `<style jsx>{` `

like on page xrp-xah-taxes.js for example.. 

so rather one or another, but don't include it in the general bundle

pandablue0809:
@ihomp 
fixed

ihomp (approved):

PR #338, opened 2025-06-13
Page with. notifications

pavel-fokin:
Suggestion for the page with notificaions. This page provides a read-only interface for listing notification rules and channels.

This PR uses "notifications", "rules", "channels" wording. I think it might be a more user-friendly and explanatory.

## Note
This PR is extensively using tailwind styles. Review if it's ok or not. I can change for a regular css modules.


ihomp:
I don't mind tailwind styles 
- if they don't mess with the rest of our UI, in case classes have similar names 
- Instead of writing CSS, ones need to learn tailwind class names - is there an easy way to design tailwind without writing classes manually? :) 
- I think we should keep "Notification" tab hidden until the time when user can create something there. Right now no new users can do anything on that tab anyway.
- I like the concept of cards, but it's a new design elements.. may be we can ask @TankaTaina to suggest css later, but for now it's ok to keep as it is. 
- We need to make sure it looks OK on mobile (where we have most of our customers) and on the dark theme.

ihomp (approved):
The code and UI looks good , but I haven't test it.

ihomp:
- We need to remove the Bots tab (commented) and the Bots page, which was a placeholder for Notifications page.

pavel-fokin:
> @pavel-fokin right now the tab Notifications doesn't give any value to teh new uers, I'm OK to merge it,
> 
> we just need
> 
> 1. disable the tab "notifications" for now, while there is no functionality there for new users.
> 2. Remove commented tab "Bots" and remove page "bots" - that was a placeholder for notifications.

IMO it's ok to have a "notifications" tab without "add notification" functionality. It serves as a teaser, we can add text "We are building notifications. Stay tuned..". We even can measure how many people click on this tab to see interest.

ihomp:
> > @pavel-fokin right now the tab Notifications doesn't give any value to teh new uers, I'm OK to merge it,
> > we just need
> > 
> > 1. disable the tab "notifications" for now, while there is no functionality there for new users.
> > 2. Remove commented tab "Bots" and remove page "bots" - that was a placeholder for notifications.
> 
> IMO it's ok to have a "notifications" tab without "add notification" functionality. It serves as a teaser, we can add text "We are building notifications. Stay tuned..". We even can measure how many people click on this tab to see interest.

We don't do teasers; we never did =) With our small team and sudden things to implement, those teasers could stack for years sometimes, and then people would not click on them when the feature went live. We've talked about notifications for years, and that's why the "Bots" tab is hidden, not left as a teaser. It needs to be removed, by the way, as you created a the notifications tab. 
No need to measure the interest, as we will implement the feature anyway, and it's already prioritized.

ihomp (approved):

Issue #332, opened 2025-06-12
New page


Anna15170221:
we need to create a new page /services/amm
Required tabs: 
AMM Create
AMM Vote

https://xrpl.org/docs/references/protocol/transactions/types/ammvote 
https://xrpl.org/docs/references/protocol/transactions/types/ammcreate

PR #339, opened 2025-06-13
New page AMM

pandablue0809:
# Issue
[we need to create a new page /services/amm](https://github.com/Bithomp/frontend-react/issues/332)


ihomp:
AMM Vote:

Asset 1 / Asset 2 - on desktop could be on the same line to save space, on mobile - 2 lines.

ihomp (changes requested):
We need fixes ass specified in comments earlier.

pandablue0809:
@ihomp 
I fixed
plz review again

ihomp (changes requested):
we need some UI fixes as described in comments

ihomp:
we also need to add 2 separate links 
Create AMM
AMM Vote 
and add them in AMM section

pandablue0809:
> 2. we need some spaces between fields when they are on the same line. in other forms, we have space between them.

what does mean?
already we have space
image

pandablue0809:
hi @ihomp 
I fixed 
plz check

ihomp:
Screenshot 2025-06-17 at 14 49 26

I see an error to chose both assets, even though they are both choosen - xrp is the default one. 

The orange messge should be as checkbox and it should be before the terms checkbox. User need to agree to it. 

"I acknowledge and understand the risks associated with XRPL AMMs, including impermanent loss, fund withdrawal risks, market volatility, and the impact of creating unbalanced pools. I agree to participate at my own risk."

AMM vote doesn't need a checkbox for terms and conditions, it's not dangerous. 
On Amm Vote we need text saying who can actually vote. 

Submit AMMVote - can be just "Vote"

ihomp (changes requested):
I see an error to choose currencies, even though both exist. in the form.

ihomp:
as currencies in the token selector are not available on testnets, let's enable links only on the mainnet for now.

ihomp (changes requested):
Wrong amounts for tokens in the signing json

PR #350, opened 2025-06-16
Add connect wallet page

pavel-fokin:
I'm experimenting with moving the `connect your wallet` flow to a separate page. Now, walletconnect and other related logic is embedded directly into each page.

This should help with:
- Reducing the app bundle size.
- Minimizing errors related to walletconnect.
- Improving overall user flow for SignIn/ConnectWallet


ihomp:
we have this screen 
Screenshot 2025-06-17 at 04 31 11

Basically, instead of having a list of wallets in the dropdown "Sign-in", we can have a "Connect a wallet" that will open that popup.

PR #351, opened 2025-06-16
Update the new explorer advantages article to fit into the learn section

pavel-fokin:
- Add new section to Learn section.
image

- Update usage of header tags.
- Generate cover image.
- Add related articles at the end.


ihomp:
The link in the footer goes to `/en/explorer-advantages`—that seems not to exist.

ihomp:
if the slug was changed, we need to update it in sitemap file too 
no need to add redirect, as probably noone bookmarked it yet

ihomp (changes requested):
The link seems broken in the footer

Issue #347, opened 2025-06-16
object is not recognised/error


Anna15170221:
Show an error when searched object is not recognised.
for example if someone search on xahau the object 0802E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C8
the api returns {"type":"unknown","error":"Could not recognise the search item"} 
but user doesn’t see any errors. we have an error field under the search field - so the error should be there

pandablue0809:
and this issue is solved in thia pr
https://github.com/Bithomp/frontend-react/pull/335

ihomp:
@pandablue0809 no, it's not solved. 

For example, try to search "0902E4E7EACC17414747174474487E9DF24CFDB61DD12C58B09D9EFB42C7F8C7" on the landing page. 

we need to handle the result 

`{"type":"unknown","error":"Could not recognize the search item"}`

PR #352, opened 2025-06-17
object is not recognised/error

pandablue0809:
# Issue
[Show an error when searched object is not recognised.](https://github.com/Bithomp/frontend-react/issues/347)


ihomp (changes requested):
Pls remove added spaces.

PR #355, opened 2025-06-17
Update tokens page

pandablue0809:
# Description
- we need a full issuer address in the list of tokens
- when searching for currencies - currencies in the dropdown should be decoded (I don’t know if we get decoded values from the backend or not, it can be decoded on the backend or frontend)
- when pressing on currency field or issuer field, it says “no options”, we should better replace it with “Start typing to search for currencies/issuers” or smth like that.. so it will be less confusing 
- scrolling for more than 100 should be available only for subscribed users
- on mobile the table should full width


ihomp:
- when currency is 3 letters, but decoded we need add (hex)
Right now, we have USD—not decoded—and USD decoded from hex, but we don't see the difference. 

- issuer field should be more like address search, but that look through issuers only. 
when I type expectar I find the address, but it doesn't say the service name. So AddressInput like input would be better.

pandablue0809:
hi @ihomp 
I fixed
plz check again

ihomp (changes requested):
Now it doesn't work to view a list by a any hex currency.

Try to type currency RLUSD and choose it.

PR #357, opened 2025-06-17
Add link to learn page and update footer layout

pavel-fokin:
This PR depends on #351.

- Add link to Learn section to the footer.
- Change footer layout as it started breaking for longer article names.

The single column, centered layout on mobile should be better for readability and UX.

## Mobile view
image

## 13 inch screen
image

### DIFF
--- a/components/Layout/Footer.js
+++ b/components/Layout/Footer.js
@@ -3,7 +3,7 @@ import { useTranslation } from 'next-i18next'
 import { useRef } from 'react'
 import dynamic from 'next/dynamic'
 
-import { devNet, nativeCurrency, network, xahauNetwork } from '../../utils'
+import { devNet, nativeCurrency, network, xahauNetwork } from '@/utils'
 
 const CookieMessage = dynamic(() => import('./CookieMessage'), { ssr: false })
 const SocialIcons = dynamic(() => import('./SocialIcons'), { ssr: false })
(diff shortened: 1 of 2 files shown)


ihomp:
In production we have it like that:
Screenshot 2025-06-17 at 18 42 45
Screenshot 2025-06-17 at 18 43 17
Screenshot 2025-06-17 at 18 42 59

I like that all social icons are in one row, I like that address is not broken in the middle and not squished—on any screen size

ihomp:
In this PR
Screenshot 2025-06-17 at 18 47 05
- all the columns do not fit in the same row on 14''
- the social icons slit into two lines 
- the text is broken a bit, like Bithomp AB - now on two different lines.. it lookes squized, even though there are space on the left. 

Screenshot 2025-06-17 at 18 46 48

ihomp:
About the mobile:
I don't mind "The single column, centered layout on mobile should be better for readability and UX.", looks good !

ihomp (changes requested):
Don't squeeze address and social icons

would be nice to fit all the columns on one line as before.. we don't have to have links for each article in the footer - if the names are long now, we can have just link to "Learn" and people can see the articles there

pavel-fokin:
@ihomp I updated width and paddings for `brand` element, so socials and address don't break. Links still colapse nicely. Pls, take a look.

ihomp:
1) It seems the PR breaks our designed layout a bit, as it changes the paddings. 
It nice to keep the same paddings as we used to have, so they stay the same everywhere. 

here what we have without PR
Screenshot 2025-06-18 at 14 42 14


here with this PR
Screenshot 2025-06-18 at 14 42 21

2. The text with company details is now align to left, not to right as it was designed. I think its nice to keep it aligned to the right as it was designed. 

3. this how it was in Russian 

Screenshot 2025-06-18 at 14 52 43

This how it is now. 

Screenshot 2025-06-18 at 14 52 28

- social icons are into lines, blocks are in two lines. 
I think the way we had it was better.  


May be we should only update how it looks on mobile, without touching it for the desktop? Otherwise we can spend quite some time for it. It's not a big issue.

ihomp:
Before
Screenshot 2025-06-18 at 14 57 41


Now
Screenshot 2025-06-18 at 14 57 54

I also liked the way before better..

centered text looks good on mobile only.

ihomp (changes requested):
May be we shouldn't touch how it looks on the desktop.. may be we can just add the link and make text centered on mobile.

pavel-fokin:
@ihomp see #370 for minimal changes

PR #359, opened 2025-06-17
Feat/improve UI object page

pandablue0809:


ihomp:
In history we see the link: Current Ledger 

I guess the "current" is not really the right word, as it's current for that time, nit current in general.
may be we can just say "Ledger index" ?

ihomp:
or does current ledger refer to the genral current ledger? why the number always satsy the same? clicking on the current ledger updates the time

ihomp (approved):

Issue #353, opened 2025-06-17
/account page updates


Anna15170221:
We need to add Tokens, Escrows, Dex trades and the last 5 the most recent transactions on /account page

ihomp:
for transactions:

```
we need to show only success transactions here - no need to show status, but we need to add a filter for successOnly in the api request
we shouldn't show transactions that were submitted by other wallets where this wallet set a regular key - that doesn't effect the current account in anyway.
time should be shown in the format like in checks .. time if today, date if more than 24h past
no need to show memos here
no need to show transaction hash, we just need an icon to transaction
if it's a payment - we need to show received or sent and then from who if received and to whom if sent.

so it should be minimum information, small section.. we will have a full list of transaction on a separate page.
```

PR #362, opened 2025-06-19
Feat/add iou data on account page

pandablue0809:
#Issue
# Description
In this PR, I implemented Token
image

### DIFF
--- a/utils/format.js
+++ b/utils/format.js
@@ -611,6 +611,7 @@ export const amountFormat = (amount, options = {}) => {
       return showValue + ' ' + valuePrefix + ' ' + textCurrency
     }
   } else {
+    //type: ['IOU', 'IOU demurraging', 'NFT']
     return showValue + ' ' + valuePrefix + ' ' + textCurrency
   }
 }


ihomp:
the data seems not updated correctly.. 
test:
enter: ihomp in the address search field and press enter
then enter bitstamp and press enter it will show 
Screenshot 2025-06-27 at 12 32 54

this is not correct data for Bitstamp... 

you need to make sure that you clear the data / update the data when address is changed. 

if you go directly to the link /en/account/Bitstamp you will see there is no tokens there. 

Though we also need to show obligations..

ihomp (changes requested):
data is not updated correctly when address is chnaged

ihomp:
change the icon for "no rippling" - to something that looks "safe", like a sheild or smth. 
no-rippling is good, that funds can't be exchanged without user approving it..

ihomp:
we also neeed to show the Limits 
may be Flags - can be updated to "Params"

and you can add an icon for limits.. when hover on it, show the limit for that trustline. - how much maximum the user can hold, until what limit he trust the another issuer.

ihomp (approved):

PR #367, opened 2025-06-19
Feat/add escrow data on account page

pandablue0809:
# Issue

# Description
In this PR, I implemented Escrow
image


ihomp:
> @pandablue0809 @ihomp I have created an account with both receiving and sending escrows: https://test.xrplexplorer.com/en/account/rU1MTfFx4Fphm941qDPDzv8nLs7Vo463Z

if you check payment channels, we have two seprate tables. 
One for sending payment channels, another one for receiving payment channels, I think we have the same for Checks. 

So, if the account is sending an escrow we show the table Outgoing escrows and show destinations. 
If the account is receiving the Escrow we need to show a separate table with SENDERS instead of destination, as destination is the current account.

pandablue0809:
image
@ihomp 
plz check

ihomp:
- to save space, we show the date or time—we have a function for that; I think it is used in issued checks. 
so, please use the same function... We should see a date or a time, not both. 

it would be nice to add escrow sequence also.. may be instead of the sequence in the first column.. or as number that showup on hover on some i -icon for information. - but it's ok to leave it for the next pr. 

DT should be before the amount, aligned to the right, both header and numbers. 

Amount column should be the last one before the actions.

ihomp:
Rename "Cancel After" to "Expire"
then instead of "doesn not cancel" -> "no expiration"

ihomp (changes requested):
Anyone showuld be able to finish escrow after the finishAfter time, anyone should be able to cancel it after the cancelAfter time. 

If there is a condition, we will need to prepare a signForm popup to enter it. For now, to make it similar, we can enable finish/cancel only for escrows without conditions. Later we can update that.

PR #370, opened 2025-06-19
Add link to Learn page in the footer

pavel-fokin:
Add link to Learn page and make minimal changes to article links.

image


ihomp:
if for seo we need 3 pages instead of one, it's ok, we can ask @Anna15170221 to split the text we had into three. 
I'll check each later and verify. 

but to save time, it would be just nice to return the page we had before. if it's not so important for SEO.

pavel-fokin:
not relevant anymore

PR #380, opened 2025-06-23
Feat/add recent transaction on accountpage

pandablue0809:
# Issue

# Description
image


ihomp:
- we need to show only success transactions here - no need to show status, but we need a filter for successOnly
- we shouldn't show transactions that were submitted by other wallets where this wallet set a regular key - that doesn't effect the current account in anyway. 
- time should be shown in the format like in checks .. time if today, date if more than 24h past 
- no need to show memos here 
- no need to show transaction has, we just need an icon to transaction 

if it's a payment - we need to show received or sent and then from who if received and to whom if sent. 

so it should be minimum information, small section.. we wi have a full list of transaction on a separate page. 

if it's all hard to do, for the first PR.. you can just make a simple table 
like time, tx status, tx type, tx link.

ihomp (commented):
Transactions should take less space.. to amke things simpler for teh first PR, we can just show timeOrDate (as in checks format), txType, Status, Link 

or we need to improve as specified in comments.

ihomp (approved):

Issue #372, opened 2025-06-19
/distribution page - new features


Anna15170221:
Add Escrow to /distribution page. 
We need the following selectable display modes:

- Without Escrow (current version) = balance
- With Escrow balances included = balance + escrow short
- With Escrow balances included = balance + escrow locked

Richlist data includes a parameter: 
`escrow`
with possible values:

`short` -  likely recipient of the escrow (usually the Destination field)
`locked `- escrow creator (should be prioritised)
`/v2/addresses/richlist?escrow=locked`

PR #384, opened 2025-06-24
/distribution page - new features

pandablue0809:
# Issue
[Add Escrow to /distribution page.](https://github.com/Bithomp/frontend-react/issues/372)


ihomp (changes requested):
we need to see the full amounts with an escrow amounts

ihomp (approved):

Issue #365, opened 2025-06-19
New Button on /tokens page


Anna15170221:
On the Page /tokens we will need a button for each token to set a trustline. 
With some default params, like 1B in amount, no rippling.

PR #398, opened 2025-06-25
New Button on /tokens page

pandablue0809:
# Issue
[On the Page /tokens we will need a button for each token to set a trustline.](https://github.com/Bithomp/frontend-react/issues/365)

# Description
- add set trustline button
- add some orders
- add service or name in token


ihomp:
- Important, rippling should be disabled when setting the trustline.. 
by default it's enabled.. that need to be changed; otherwise users can lose their funds.

ihomp:
Rename set trustline on the button to set trust, or replace it with a small icon with the hover "Set trust"

ihomp:
also add class narrow to the set trust button

ihomp (approved):

PR #402, opened 2025-06-25
Add tabs Withdraw and Deposit to AMM

pandablue0809:
# Issue

https://github.com/Bithomp/frontend-react/issues/396


ihomp:
if we have a half length components, they need to be alligned
Screenshot 2025-06-27 at 18 43 27
Screenshot 2025-06-27 at 18 43 17
Screenshot 2025-06-27 at 18 43 08

ihomp:
We need the same checkbox as in the AMM create

ihomp:
The order for the tabs and the menu should be: 

- Amm deposit 
- Amm withdrawal 
- Amm vote 
- Amm create

ihomp (commented):
Great work, but let's improve the UI a bit

pandablue0809:
@ihomp 
I fixed
plz check again 🙏

ihomp:
- The code looks a bit bloated - as many pieces are repeated for each separate case.. 
- it's ok to leave it as it is, for simplicity and in order not to spend too much time for it. 

- I haven't tested it yet.
- @Anna15170221, can you please test this one locally? 
- make sure that "Trading Fee Vote" is correct. 
- Also suggest if any text updates and additions are needed, or links/explanations.

ihomp (approved):

PR #421, opened 2025-07-01
Update Card design

pavel-fokin:
Follow-up on Slack thread, suggestion to update Card design. @ihomp What do you think? 

- Added CSS variables for the color system to be used in the design. It's better to have a consistent scale to use for different elements.
- Updated the Card component.

## Learn page
image

## Notifications admin page
image


pavel-fokin:
We can add icons on Learn page making it more vivid. In next PR
image

Issue #407, opened 2025-06-27
/send page fixes needed


Anna15170221:
On page send payment, when I type address like Bithomp and then choose that in the dropdown, the address is filled. but the username/service name is not on top of the component. it’s there for a second. but then disappears - it would be nice fix that. So the username/service name stays on top of the component when the address is chosen.

PR #429, opened 2025-07-01
/send page fixes needed

pandablue0809:
# Issue
[On page send payment, when I type address like Bithomp and then choose that in the dropdown, the address is filled. but the username/service name is not on top of the component. it’s there for a second. but then disappears - it would be nice fix that. So the username/service name stays on top of the component when the address is chosen.](https://github.com/Bithomp/frontend-react/issues/407)


ihomp:
We have the same issue for many other pages like in /faucet 

the problem here that the component re-renders when the address changes.. 
the addressDetails are erased when the component addressInput re-renders.. 
and the addressDetails never leaving the addressInput component... 

my seggestion
if we can't get away from component re-render, then we can probably set addressDetails from the addressInput as addressInput already have this data it just doesnt pass it to the parent page.

ihomp (changes requested):
We rather need a universal solution that would work for all other cases as well.

pandablue0809:
I fixed @ihomp

ihomp (changes requested):
This is over-complecated.. 
- components requests data, pass it to the page, then page pass it back to a component, when the data is only used inside of the component. 
- you've added the same api request on the page, that already in use in the component, when data is only used on the componnet. 

we have a working solution: 
https://test.xrplexplorer.com/en/faucet

- When you specify the address, the username/service name will stay. 
- it's good to trigger add the username/service name for the default value when the value is specified in the url (but within the component)

ihomp:
fixed in https://github.com/Bithomp/frontend-react/commit/ac89189ebd07033dee19a938859c3453779a7c07

Issue #408, opened 2025-06-27
/tokens page improvements needed


Anna15170221:
We will need some verification status for tokens, and a filter to show only verified tokens.

![Image](https://github.com/user-attachments/assets/a2594c63-1c50-4f26-b10f-aa61502352bb)

We need to add here the trustlines count and holders count.
We need to fit it on the top line preferable, but can be on a separate line

PR #433, opened 2025-07-01
/tokenSelector improvements needed

pandablue0809:
# Issue
[We need to add here the trustlines count and holders count.](https://github.com/Bithomp/frontend-react/issues/408)


ihomp (changes requested):
we need some small ui fixes

pandablue0809:
@ihomp 
I fixed

Issue #411, opened 2025-06-27
checks need to be checked :)


Anna15170221:
checks like (currency === nativeCurrency)— that’s not always XRP
you also need to check that there is no issuer, because there are many fake xrp, that someone could issue, but all of them would have an issuer. the real xrp won’t have an issuer.
so please check where you have such checks and fix it by checking also the issuer. 
you can add a function in utils, and use it for checking realxrp or not.

PR #435, opened 2025-07-01
checks need to be checked

pandablue0809:
# Issue
[checks like (currency === nativeCurrency)— that’s not always XRP](https://github.com/Bithomp/frontend-react/issues/411)


utils/index.js
```diff
@@ -785,6 +785,19 @@ export const isAmountInNativeCurrency = (amount) => {
   return !amount?.issuer && !amount?.mpt_issuance_id
 }
 
+export const isNativeCurrency = (currencyObj) => {
+  if (!currencyObj) return false
```
ihomp, line 789: When it is a number as a string, not an object, it also can be a native currency. 
we should probably add this case too.

pandablue0809:
@ihomp 
I fixed

ihomp (approved):

ihomp (approved):

Issue #418, opened 2025-06-30
/nft-explorer new feature - search on sale for any token


Anna15170221:
for /nft-explorer in the filters we have onSale for XRP/XAH only
we need to replace it with token selector, so user can see nfts on sale for any token

PR #438, opened 2025-07-01
/nft-explorer new feature - search on sale for any token

pandablue0809:
# Issue
[we need to replace it with token selector, so user can see nfts on sale for any token](https://github.com/Bithomp/frontend-react/issues/418)


components/NftsComponent.js
```diff
@@ -121,7 +137,7 @@ export default function NftsComponent({
 
   const listTabList = [
     { value: 'nfts', label: t('tabs.all') },
-    { value: 'onSale', label: t('tabs.onSale', { nativeCurrency }) }
```
ihomp, line 124: instead of removing the usage of` t('tabs.onSale')` we need to update it in all languages. 
we can just make it as "Listed". 

`t('tabs.onSale', { nativeCurrency })` will become `t('tabs.onSale')`

components/NftsComponent.js
```diff
@@ -638,7 +677,7 @@ export default function NftsComponent({
             (isValidTaxon(taxonQuery) ? ' ' + taxonQuery : '') +
             (ownerQuery ? ', ' + t('table.owner') + ': ' + ownerQuery : '') +
             (activeView === 'list' ? ' ' + t('tabs.list') : '') +
-            (listTab === 'onSale' ? ' ' + t('tabs.onSale', { nativeCurrency }) : '') +
+            (listTab === 'onSale' ? ' ' + 'On sale for tokens': '') +
```
ihomp, line 680: let's use `t('tabs.onSale')`

components/NftsComponent.js
```diff
@@ -708,7 +747,16 @@ export default function NftsComponent({
...
+              {nftExplorer && (
+                <div>
+                  {t('table.currency')}
+                  <TokenSelector
+                    value={selectedToken}
+                    onChange={setSelectedToken}
+                  />
+                </div>
+              )}
+              {!nftExplorer && saleCurrencyIssuer && saleCurrency && (
```
ihomp, line 759: why would we make it different for !nftExplorer? 
we should keep the logic as we had it without a separation. 

do for /nfts/ /nft-explorer it will be the same

ihomp (changes requested):
view the comments

pandablue0809:
@ihomp 
fixed

Issue #422, opened 2025-07-01
/distribution page - improvements needed


Anna15170221:
For the tables with escrows we need to show the summ of both balance+escrow in the third column.

PR #440, opened 2025-07-02
/distribution page - improvements needed

pandablue0809:
# Issue
[For the tables with escrows we need to show the summ of both balance+escrow in the third column.](https://github.com/Bithomp/frontend-react/issues/422)


ihomp (changes requested):
We need refactoring here. 
We shouldn't copy-paste the same code in two different locations, it should be extracted as small-component in the same file or as function.

Issue #423, opened 2025-07-01
Timemachine API calls


Anna15170221:
When using a timemachine, the react sends two API calls to transactions. We should have only one call. Kindly fix it.

PR #441, opened 2025-07-02
Timemachine API calls

pandablue0809:
# Issue
[When using a timemachine, the react sends two API calls to transactions. We should have only one call. Kindly fix it.
](https://github.com/Bithomp/frontend-react/issues/423)


ihomp (commented):
Moving the fetch to the parent doesn't look right. It's better to keep it inside of the component for an easier maintance. 

We need to find out what is causing the rerender of the component or page. That can be some passed variables, like a flat rate, etc. so that probably needs to be fetched on the server side, before rendering.

Issue #432, opened 2025-07-01
We need to add conversion of inputs longer than 3 letters into hex. /trustline page - currency code field


Anna15170221:
when the user enters more than three letters, the input should be converted to hex (if the user didn't enter hex originally).

PR #444, opened 2025-07-02
We need to add conversion of inputs longer than 3 letters into hex. /trustline page - currency code field

pandablue0809:
# Issue
[when the user enters more than three letters, the input should be converted to hex (if the user didn't enter hex originally).
](https://github.com/Bithomp/frontend-react/issues/432)


ihomp (changes requested):
we should rather limit how many characters user can enter as a currency name that is not HEX, instead of truncating the value. 

We can not truncate—that means we will send for transaction something different from what user enters.

pages/services/trustline.js
```diff
@@ -119,6 +120,11 @@ export default function TrustSet({ setSignRequest }) {
       return
     }
 
+    if (currency.currency.length > 3 && !/^[0-9A-F]{40}$/i.test(currency.currency)) {
```
ihomp, line 123: We need to allow the enter any text up to 20 characters or hex 40. 

The error can be 'Please enter a valid currency code, eg USD, MyCurrency or HEX currency code (40 characters).'

pages/services/trustline.js
```diff
@@ -259,13 +265,53 @@ export default function TrustSet({ setSignRequest }) {
...
+                            setCurrency({ currency: hexCurrency })
+                          }
+                        }}
+                      >
+                        Convert
+                      </div>
+                    </div>
+                  </div>
+                  <span className="orange">
+                    <b>Note:</b> If currency more than 3 characters, it should be converted to HEX. Max 20 characters for non-HEX codes.
```
ihomp, line 311: We don't need this NOTE, we should convert to hex ourselves if it's not hex.

ihomp (changes requested):
we need fixes as in comments

pages/services/trustline.js
pandablue0809: what does mean?
I think If currency's length >3, we should convert hex
so I think this line is correct

ihomp: We don't need the convert button, we should do it on the fly for currencies over 3 characters and less than 20, if it isn't a 40 character hex.

ihomp: no need to have a note "If currency more than 3 characters, it should be converted to HEX. " we should just do it.. 

if its 3 characters - leave it as it is, if it's less than 3 it's invalid.. if it more than 20 and not equal 40 (and not hex) - it's invalid 

if it is from 3 to 20 - convert automatiocally to hex

ihomp: correct

ihomp (approved):

Issue #461, opened 2025-07-14
currencyDetails=true for displaying LP tokens in token list and search


Anna15170221:
Please update the token list and token search features to use the currencyDetails=true parameter when fetching token data. This is necessary to properly display LP tokens.
The response format will change when using currencyDetails=true, so some adjustments will be needed.
This update should be applied both in the token list and in the token search functionality.

ihomp:
I’m testing the LP tokens in tokens and currency search. There are several issues: 
if I type in currency search “LP” it takes very long time, and then returns lot’s of results. I think we should limit it by 20, and then show the message to be more specific - the same message as we have in the token search. 
when I choose an LP token - the field get prefilled with the currency code 
as we have the currency code in the field, it’s nice to have the in the title (next to the word currency) the actual currency in the human format, like Currency: XRP/BTC (Bitstamp) - yes we need to specify the issuer here.. otherwise it’s hard to recognise. 
3. we need to add the currency code and the issuer in the URL, so users will have direct links for the views they need. 
4. marketcap someties show just “XRP”, we need to make it 0 XRP, or “no market” 
5. in the token list, we need to show the issuers also.. without isusers it’s hard to differentiate 
so like BTC (Bitstamp) / Coffee (rwerr…werwer) 

<img width="1381" height="573" alt="Image" src="https://github.com/user-attachments/assets/9b520f5a-a970-40ce-8b19-489dd31c56cf" />

PR #467, opened 2025-07-15
currencyDetails=true for displaying LP tokens in token list and search

pandablue0809:
# Issue
[update the token list and token search features to use the currencyDetails=true parameter when fetching token data. This is necessary to properly display LP tokens](https://github.com/Bithomp/frontend-react/issues/461).


ihomp (changes requested):
We need to add it only when it's in use to show LP tokens properly.. 
in the token search and in the token list.. 

Tokens should be named like 
XRP/RLUSD LP instead of just LP Token

Issue #460, opened 2025-07-14
Escrows - avoid unnecessary api calls


Anna15170221:
we have a functionality in escrows - when user clicks on the cancel/execute we request the escrow data, get the sequence number and then create transaction. 
Now escrow objects by default will contain escrowSequence   - so we don’t need to request in additional api the sequence number, we will have it right away.
Kindly check and rewrite that part, so we don’t do unnecessary api calls.

Issue #458, opened 2025-07-11
/account page - escrows view


Anna15170221:
Show ONLY available actions or “none” in the actions column.

PR #468, opened 2025-07-16
Escrows - avoid unnecessary api calls

pandablue0809:
# Issue
- [we don’t do unnecessary api calls.](https://github.com/Bithomp/frontend-react/issues/460)
- [Show ONLY available actions or “none” in the actions column.](https://github.com/Bithomp/frontend-react/issues/458)


ihomp:
It seems the `escrowSequence` is not available... I can't test. @ildaruz 
`/en/account/wietse` the `escrowSequence` is `null` in objects / escrows

ihomp (approved):

ihomp:
@ildaruz shall we keep the old behavier as a backup (when we send an aditional api call to request the sequence) or are we confident that soon it will be always available?

ildaruz:
> @ildaruz shall we keep the old behavier as a backup (when we send an aditional api call to request the sequence) or are we confident that soon it will be always available?

I recomend to keep it for backup, in case crawler is outdated and we still need it untill sequence is fullfilled

Issue #452, opened 2025-07-10
pop-up to sign-in with email


Anna15170221:
We need to create a pop-up to login with email to https://bithomp.com/en/admin.
When we don't leave the current page but authorize in the popup (like we do with the wallets).
It will be used on the pages where functionality is limited by PRO subscription (for example advanced options on /check, /send or /escrow). 
With popup
user doesn't leave his current page
user won't need to fill in the fields again.

PR #470, opened 2025-07-16
pop-up to sign-in with email

pandablue0809:
# Issue
[We need to create a pop-up to login with email to https://bithomp.com/en/admin.](https://github.com/Bithomp/frontend-react/issues/452)


pandablue0809:
> It seems that this PR also includes commits/changes related for DelegateSet, which shouldn't be the case.

I fixed @ihomp 
plz check again 🙏

Issue #406, opened 2025-06-27
/nft-mint and /nft pages - sell offer with token


Anna15170221:
We need to limit the sell offer token select currencies only to the ones that user has trustlines for with limit over 0.

PR #474, opened 2025-07-23
/nft-mint and /nft pages - sell offer with token

pandablue0809:
# Issue
[We need to limit the sell offer token select currencies only to the ones that user has trustlines for with limit over 0.
](https://github.com/Bithomp/frontend-react/issues/406)


ihomp:
@pandablue0809 we need the same for the URITokenMint ...

ihomp:
for NFTokenCreateOffer the destination would probably depends if it's the buy or sell offer... 

for the sell offer, yes, the "destination" is the seller - the one who is selling. 

for the buy offers - destination is the nft "owner"

pandablue0809:
@ihomp 
I fixed
plz review again 🙏

Issue #448, opened 2025-07-04
live price updates


Anna15170221:
We need to implement live price updates in the system so that:

The calculator on the main page reflects real-time rates

ws subscribe to rates stream
`{ command: "subscribe", streams: ["rates"], currency: "usd", id: n }`

ws unsubscribe from all or single currency in rates stream
```
{ command: "unsubscribe", streams: ["rates"], id: n }
{ command: "unsubscribe", streams: ["rates"], currency: "usd", id: n }
```
example message

`{"usd":2.16402,"type":"rates"}`

PR #491, opened 2025-07-24
live price updates

pandablue0809:
# Issue
[We need to implement live price updates in the system](https://github.com/Bithomp/frontend-react/issues/448)

### DIFF
--- a/hooks/useRates.js
+++ b/hooks/useRates.js
@@ -1,168 +1,68 @@
-import { useState, useEffect, useRef } from 'react'
+import { useState, useEffect } from 'react'
 import { wssServer } from '../utils'
 import { fetchCurrentFiatRate } from '../utils/common'
 
+let ws = null
 let requestId = 1
+
+function sendData(currency) {
+  if (ws.readyState) {
+    ws.send(JSON.stringify({ 
+      command: 'subscribe', 
+      streams: ['rates'], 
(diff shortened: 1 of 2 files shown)


ihomp:
1) We used query the fiatRate only when the user changes the fiat currency or enters a page that uses the fiatPrice 

`const allowedRoutes = ['/', '/account/[[...id]]', '/amms', '/distribution', '/admin/watchlist', '/tokens']`

[`selectedCurrency, router.pathname])`

Now it seems that we will connect to a websocket even if we open the page which doesnt use the fiatPrice. 

2) second thing, we already have some other connections to the same websocket.. for statistics and for the whales on the landing page. 

3) we don't need Live fiat prices for the rest of the pages - that will cause re-renders on pages / components. 

----------//------------
As we need live price only for the landing page for now. lets don't change functionality for other pages. 
we need to check the way how to use only one websocket for all streems, like whales, stats and price.. so it will be only one connection instead of three.. and we should not open new websockets when we go to nft statistics or ledger data etc pages that also use web socket. Websocket should be one for the whole app.

pandablue0809:
@ihomp 
I fixed

ihomp:
- Does it reconnect after WS closing on error? I think it doesn't. 
- Please solve the conflict

ihomp (approved):

Issue #445, opened 2025-07-04
transactions - new page


Anna15170221:
We need to create a new page to show all transactions of the wallet.
The same info and search options that we have here in the column Transactions https://bithomp.com/explorer/rfjquSFSKXnDYdQyGWuXBSSSW87jfpX7Qm
Ask @ihomp to provide php code to create the similar one.

ihomp:
I think we need to use v3/transactions so the data will be enriched with accountDetails for each address. so we would be able to show usernames and service names without any additional api calls.

PR #498, opened 2025-07-28
transactions - new page

pandablue0809:
# Issue
[We need to create a new page to show all transactions of the wallet.
](https://github.com/Bithomp/frontend-react/issues/445)


components/Layout/SearchBlock.js
```diff
@@ -298,6 +298,11 @@ export default function SearchBlock({ searchPlaceholderText, tab = null, userDat
         return
       }
 
+      if (tab === 'transactions') {
+        router.push('/transactions/' + encodeURI(searchFor) + addParams)
```
ihomp, line 302: we don't need that - othewise when users search for address in /transactions 

address won't work.. no?

components/UI/TransactionBlock.js
```diff
@@ -0,0 +1,506 @@
...
+          </span>
+        )}
+      </span>
+    </span>
+  </span>
+);
+
+const TransactionBlock = ({ tx, index, address, isMobile }) => {
+  if (!tx) return null;
+  const dateText = dateFormat(tx.date + 946684800) || '-';
```
ihomp, line 39: we should better use utils, don't we already have converters for ripple epoch?

components/UI/TransactionBlock.js
```diff
@@ -0,0 +1,506 @@
...
+                      {shortAddress(nft.index)}
+                    </Link>
+                    <br />
+                </>
+              )}
+            </div>
+          ))}
+        </>
+      );
+    } else if (tx.nftType === 'nftokenacceptoffer') {
```
ihomp, line 110: We should have different types in separate components, like we did for /transaction/. that way it will be easier to maintain.

utils/transactionBlock.js
```diff
@@ -0,0 +1,916 @@
...
+import { nativeCurrency } from '.';
+
+export function txTypeToText(type, capitalize = false) {
+  if (!type || typeof type !== 'string') {
+    type = 'unknown';
+  }
+
+  let formattedType = '';
+  
+  switch (type.toLowerCase()) {
```
ihomp, line 11: One more type conversion? =))

ihomp:
Design I suppose was taken from here https://www.figma.com/proto/RSKPpB13mzBAqlfPHXG6lq/Bithomp?node-id=732-2740&t=avN7B3zZEPPO7mjU-1&scaling=scale-down&content-scaling=fixed&page-id=360%3A2&starting-point-node-id=732%3A2395&show-proto-sidebar=1 

we would need to add Fiat values as well there

components/Layout/SearchBlock.js
pandablue0809: I think  it's need
I think if we seach address in transactions page, we have to go this address's transactioins page, not address page

components/UI/TransactionBlock.js
ihomp: we should add it then

ihomp:
```
pages/
└── account/
    ├── index.js                      ← handles /account
    ├── [id].js                  ← handles /account/{address}
    └── [id]/
        └── transactions.js           ← handles /account/{address}/transactions
```

This is the structure for the routes.
1) we need to replace `/account/[[...id]].js` with two pages, `/account/index.js`, when no address is specified. and `/account/[id].js ` when there is an address 
2) we need to add `/account/[id]/transactions.js` with the transactions functionality.

components/Account/RecentTransactions.js
```diff
@@ -48,7 +48,7 @@ export default function RecentTransactions({ userData, ledgerTimestamp }) {
         <thead>
           <tr>
             <th colSpan="100">
-              Last 5 transactions [<a href={'/explorer/' + address}>View all</a>]{historicalTitle}
+              Last 5 transactions [<a href={'/account/' + address + '/transactions'}>View all</a>]{historicalTitle}
```
ihomp, line 51: It's too early to add the links.. 

so, revert the chnages here for now..

when we add the links, it will be <Link> not <a>,ha so they will have the language code as well.

components/Account/RecentTransactions.js
```diff
@@ -98,7 +98,7 @@ export default function RecentTransactions({ userData, ledgerTimestamp }) {
       <div className="show-on-small-w800">
         <br />
         <center>
-          {'Last 5 Transactions'.toUpperCase()} [<a href={'/explorer/' + address}>View all</a>]{historicalTitle}
+          {'Last 5 Transactions'.toUpperCase()} [<a href={'/account/' + address + '/transactions'}>View all</a>]{historicalTitle}
```
ihomp, line 101: too early to add links, and we will need <Link> for intrenal react links

pages/account/[id].js
```diff
@@ -19,7 +19,7 @@ const setBalancesFunction = (networkInfo, data) => {
   if (!data?.ledgerInfo || !networkInfo) return null
   let balanceList = {
     total: {
-      native: data.ledgerInfo.balance
+      native: data.ledgerInfo.balance || 0
```
Comment (ihomp) [inline pages/account/[id].js:22]: why would you add 0 here? no body want to see 0 when there is a network error or smth similiar, they should better see an error rather than balance as 0.

pages/account/[id].js
```diff
@@ -120,7 +120,7 @@ export default function Account({
   fiatRateServer,
   selectedCurrencyServer,
   networkInfo,
-  balanceListServer
+  balanceListServer,
```
Comment (ihomp) [inline pages/account/[id].js:123]: do you have different rules for js prettify? we don;'t add commas in the end of objects

pages/account/[id].js
```diff
@@ -169,15 +162,12 @@ export default function Account({
       if (initialData.obligations?.trustlines > 200) {
         setGateway(true)
       } else {
-        //keep it here for cases when address changes without refreshing the page
         setGateway(false)
       }
     } else {
-      //keep it here for cases when address changes without refreshing the page
```
Comment (ihomp) [inline pages/account/[id].js:176]: please don't remove comments

pages/account/[id].js
```diff
@@ -169,15 +162,12 @@ export default function Account({
...
       setGateway(false)
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [initialData])
+  }, [initialData, setActivatedAccount])
```
Comment (ihomp) [inline pages/account/[id].js:170]: are you sure you want to add setActivatedAccount here?

pages/account/[id].js
```diff
@@ -508,7 +496,6 @@ export default function Account({
                             <tbody>
                               <tr>
                                 <td colSpan="2" className="no-padding">
-                                  {/*  (info) Check account balance and settings in any Time in the past. */}
```
Comment (ihomp) [inline pages/account/[id].js:511]: pls do not remove the comments

styles/components/dropdown.scss
```diff
@@ -14,6 +14,12 @@ $small-screen: 440px; // minimum screen
     }
   }
 
+  &--filters {
```
ihomp, line 17: what this for? how to test it?

ihomp (commented):
Please return the comments and answer some questions.

styles/components/dropdown.scss
pandablue0809: it's need
simpleSelect is disply hidden in mobile view
so we have to add css

components/Layout/SearchBlock.js
```diff
@@ -337,6 +342,8 @@ export default function SearchBlock({ searchPlaceholderText, tab = null, userDat
       ['amm', 'account', 'nft', 'nfts', 'nft-offer', 'nft-offers', 'transaction', 'nft-volumes', 'object'].includes(tab)
     ) {
       return t('explorer.header.' + tab)
+    } else if (tab === 'transactions') {
+      return t('Transactions')
```
ihomp, line 346: That looks invalid, as we do not have such a translation piece in the translation files.

Keep it just "Transactions," or add the translation to all 8 files. 

Actually, we have already translated

`t("explorer.menu.transactions")`

ihomp:
```
const [pageFiatRate, setPageFiatRate] = useState(0)

  useEffect(() => {
    if (!selectedCurrency || !tx?.outcome) return
    const { ledgerTimestamp } = tx?.outcome
    if (!ledgerTimestamp) return

    fetchHistoricalRate({ timestamp: ledgerTimestamp * 1000, selectedCurrency, setPageFiatRate })
  }, [selectedCurrency, tx])
```

This code is repeated in many transaction types, as many of them need the historical price.  (we shouldn't not copy-paste the code to different places)
You should better subtract it from each tx Card and include it in the Card , it already has the `tx` and `selectedCurrency`
then `pageFiatRate` can be passed into tx cards.

components/Transactions/TransactionRowAccountDelete.js
```diff
@@ -0,0 +1,59 @@
+import { TransactionRowCard } from './TransactionRowCard'
+import { useEffect, useState } from 'react'
+import { fetchHistoricalRate } from '../../utils/common'
+import { addressUsernameOrServiceLink, amountFormat, nativeCurrencyToFiat } from '../../utils/format'
+import { FiDownload, FiUpload } from 'react-icons/fi'
+
+
```
ihomp, line 7: Don't you have a prettify package in your code editor? The second empty line is unnecessary.

pages/account/[id].js
Comment (ihomp) [reply pages/account/[id].js:22]: so, do we need` || 0 `here ?

pandablue0809:
@ihomp 
I fixed

Issue #459, opened 2025-07-14
bug in the token select on /send page


Anna15170221:
https://bithomp.com/en/services/send?address=rNxp4h8apvRis6mJf9Sh8C6iRxfrDWN7AV
when destination is entered: instead of showing ONLY the tokens that account can hold we show rubbish trustlines that were set by third parties to the destination address.
We don’t show those tokens in the https://bithomp.com/en/address/rNxp4h8apvRis6mJf9Sh8C6iRxfrDWN7AV, so we shouldn’t have them in the token select neither 

<img width="1008" height="1130" alt="Image" src="https://github.com/user-attachments/assets/74e59276-0abe-49fe-abfa-bb3e419d81f3" />

PR #499, opened 2025-07-29
bug in the token select on /send page

pandablue0809:
# Issue
[when destination is entered: instead of showing ONLY the tokens that account can hold we show rubbish trustlines that were set by third parties to the destination address.](https://github.com/Bithomp/frontend-react/issues/459)

### DIFF
--- a/components/UI/TokenSelector.js
+++ b/components/UI/TokenSelector.js
@@ -21,21 +21,21 @@ const limit = 20
 // Helper function to fetch and process trustlines for a destination address
 const fetchTrustlinesForDestination = async (destinationAddress, searchQuery = '') => {
   const response = await axios(`v2/trustlines/tokens?trustline=${destinationAddress}&limit=${limit}`)
-  const objects = response.data?.objects || []
-
-  // Filter RippleState objects to get trustlines where destination can hold tokens
-  const trustlines = objects.filter((obj) => {
-    if (parseFloat(obj.LowLimit.value) <= 0 && parseFloat(obj.HighLimit.value) <= 0) return false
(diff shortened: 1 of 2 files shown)


ihomp (changes requested):
Lets address my commnets in the ticket.
1 - it still loads uncreated trustlines 
2 - updates the list on opening, when no params changed.

pandablue0809:
@ihomp 
I fixed reload issue

ihomp:
on the send page 

http://localhost:3000/en/services/send

if I chose a currency for example RLUSD
- close the token selector 
- re-open token selector -> it will show the loading state - even though nothing is changed ... we shouldn't have data reloads when search is not changed.

pandablue0809:
@ihomp 
fixed

ihomp:
Let's wait for the new endpoint by @ildaruz 

`v2/trustlines/tokens/destination/${destinationAddress}?limit=${limit}`

ihomp:
@pandablue0809 please update the endpoint to use 

`/v2/address/:address/acceptedTokens`

ihomp:
priority 0 @Anna15170221

pandablue0809:
@ihomp 
I fixed

ihomp:
@ildaruz the data seems invalid, we only need the list of tokens, not all the trustlines

https://bithomp.com/api/v2/address/rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B/acceptedTokens?limit=20

priority 0. 

plus ad the nativeToken: XRP/XAHAU

ihomp (approved):

PR #500, opened 2025-07-30
fix style amendment page

pandablue0809:
# Issue
[we don't need to show fields in the table that are undefined, because there are no such for already activated or obsolete](https://github.com/Bithomp/frontend-react/issues/313)

### DIFF
--- a/pages/amendment/[amendmentName].js
+++ b/pages/amendment/[amendmentName].js
@@ -123,18 +123,18 @@ export default function AmendmentSummary({
     }
   }, [validators, amendmentData])
 
-  const amendmentId = amendmentData?.amendment || '-'
-  const introduced = amendmentData?.introduced || '-'
-  let threshold = '-'
+  const amendmentId = amendmentData?.amendment
+  const introduced = amendmentData?.introduced
+  let threshold = ''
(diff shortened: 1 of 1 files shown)


pandablue0809:
@ihomp 
I fixed 
plz review again 🙏

ihomp:
The new page is missing any links back to amendments. 

1)
- Let's add the amendment tabs (same as on the amendments page)—without adding a new tab.
- so it will be [validators, amendments, votes]—and all of them unselected 

2) http://localhost:3000/en/amendment/fixInvalidTxFlags

This page says: This page shows validator votes for the fixInvalidTxFlags amendment on the XRP Ledger.

But in fact, it doesn't show the votes. 

Rather, change the text to "This page displays information for the fixInvalidTxFlags amendment on the XRP Ledger," or remove the text, or add different text on pages depending on whether it is enabled or not. 

3) http://localhost:3000/en/amendment/fixNFTokenDirV1
This is an obsolete amendment. 

But on the page it doesn't say that. 
For them, we only need to show the status that is the obsolete amendment. 
Name, ID, and when introduced—the rest is irrelevant and should be omitted.

ihomp (changes requested):
we need some more small UX updates.

Issue #508, opened 2025-07-31
/trustline - show token description


Anna15170221:
/services/trustline

1. when selecting a token, display the token description and other related information – we already have this data in our database.

2. Add the warning: 
We do not take responsibility for the accuracy of the token descriptions or related information. Users should always do their own research (DYOR). The content is for informational purposes only, not financial advice.

PR #514, opened 2025-08-01
/trustline - show token description

pandablue0809:
# Issue
[when selecting a token, display the token description and other related information – we already have this data in our database.
](https://github.com/Bithomp/frontend-react/issues/508)

### DIFF
--- a/pages/services/trustline.js
+++ b/pages/services/trustline.js
@@ -276,8 +276,11 @@ export default function TrustSet({ setSignRequest, currencyQuery, currencyIssuer
                 currencyQueryName="currency"
               />
               {selectedToken.description && (
-                <div>
-                  <span className="grey">{selectedToken.description}</span>
+                <div style={{ marginTop: 10 }}>
+                  <span className="grey">
+                    <b>Description:</b> {selectedToken.description}
+                  </span>
+                  <br />
                   <br />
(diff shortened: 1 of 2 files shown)


pandablue0809:
> TokenSelector functionality is not what is expacted
> 
> 
> open http://localhost:3000/en/services/send
> 
> * open token selector
> * type xrp -> the search will find xrp names coins
> * erase xrp from search, it will stay the same
> * even if you close and open the search is epty, but the search is not updated to the original one, when ripple RLUSD chould be the first.
> 
> 2. if you type bitstamp in toen selector -< it will find bitstamp tokens
> 
> * then close the token selector
> * and open again token selector - it will re-load the list again, even though the search is not changed.

I will fix all issue about tokenselector in this [PR](https://github.com/Bithomp/frontend-react/pull/499)

pages/services/trustline.js
```diff
@@ -217,6 +217,7 @@ export default function TrustSet({ setSignRequest, currencyQuery, currencyIssuer
       setError(err.message)
     }
   }
+  console.log(selectedToken)
```
ihomp, line 220: Please don't leave the `console.log` in the PRs.

ihomp (approved):

Issue #505, opened 2025-07-31
/token/{issuer}/{currencyCode} -New page


Anna15170221:
We need to create a new page to display the full statistics and historical data
/token/{issuer}/{currencyCode} 
So this will be a separate page for every single token which displays the full data in one place. 
Priority 1.

PR #524, opened 2025-08-08
/token/{issuer}/{currencyCode} -New page

pandablue0809:
# Issue
[We need to create a new page to display the full statistics and historical data
](https://github.com/Bithomp/frontend-react/issues/505)

### DIFF
--- a/pages/token/[[...id]].js
+++ b/pages/token/[[...id]].js
@@ -1,0 +1,439 @@
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { useEffect } from 'react'
+import { Link, useRouter } from 'next/router'
+
+import SEO from '../../components/SEO'
+import { TData } from '../../components/Table'
+import {
+  capitalize,
+  niceNumber,
+  shortNiceNumber,
+  fullNiceNumber,
+  CurrencyWithIcon,
+  AddressWithIconFilled
+} from '../../utils/format'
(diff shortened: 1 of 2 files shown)


ihomp:
We should better do it in a format like an NFT page, where the icon is big on the left side.
Then we will have buttons under the big icon like 
"Set trustline" 

On the right side, there will be different sections. 

Like "Token information" "Stats for the last 24h", "Stats for the last closed day". 
so we need to split data into sections, depending on the time when it was last updated... 

Also we need to show all the information we have; there is no need to hide xrp-price changes etc.

ihomp (changes requested):
1. on the left side we need to show icon of the currency not the avatar of the address (on the tokens page we show already token images) /{issuer}/{currency} 
2. the number like supply, holders, trustlines - lets show right away in full format, no need to shorten it, as we have enough space here, no need for tooltips probably.. 
3. price informations seems calculated wrong.. the prices are in xrp, so you need to get hostorical xrp prices in fiat in order to calculate it right, you can't multiply by the current rate to get historical prices. 
4. price information format, must be in one line, like 0.9999 USD (0.23 XRP) 
5. same for volumes total, they also should in one line and with mode detaled numbers like 12587.65 IOU (100098.34 USD)

In general, we need to show the numbers in more detailed. Some numbers are in wrong sections, but we can fix it later.

Issue #515, opened 2025-08-02
landing page - live fiat prices in the chart


Anna15170221:
Let’s push the live fiat prices to the chart. so it’s time + price, so the graph will update/grow live.

ihomp:
@Anna15170221 it's deployed, please test, close, and announce.

PR #525, opened 2025-08-08
landing page - live fiat prices in the chart

pandablue0809:
# Issue
[Let’s push the live fiat prices to the chart. so it’s time + price, so the graph will update/grow live.
](https://github.com/Bithomp/frontend-react/issues/515)

### DIFF
--- a/components/Home/PriceChart.js
+++ b/components/Home/PriceChart.js
@@ -344,6 +344,7 @@ export default function PriceChart({ currency, chartPeriod, setChartPeriod, hide
     }
 
     const bucketMs = getBucketMs(chartPeriod)
+    const bucketStart = (timestamp) => Math.floor(timestamp / bucketMs) * bucketMs
 
     setData((previousData) => {
       const previousSeries = Array.isArray(previousData?.data) ? previousData.data : []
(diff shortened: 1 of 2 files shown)


ihomp (changes requested):
If I leave the default chart I don't see new points..

for example open the chart at 12:29

12:00
12:29

then it will 

12:00
12:30

then 

12:00
12:31

but should be 

12:00
12:30 - this is missing 
12:31

same if you open chart at 12:44

12:00
12:44

then t should be 

12:00
12:45 - this is missing 
12:46

ihomp:
right now if I open a chart at 10:29 
I'll have the data points on the chart with those times:
10:00
10:29

at 10:30, it will be points 
10:00
10:29
10:30 

at 19:31, it will be points 

10:00
10:29
10:31 

BUT It should be at 10:30
10:00
10:30

and at 10:31 it should be 
10:00
10:30
10:31

ihomp:
@Anna15170221 we need to wait for this ticket before anouncment.

ihomp (changes requested):
data points added at wrong times

Issue #493, opened 2025-07-28
services/trustline - share button


Anna15170221:
Add a "share" button that automatically selects the current field settings in the browser link using query parameters.

ihomp:
We need a share button for services/trustline. 

That will copy the current URL to set up the trustline

moonfury-ops:
Hi @ihomp 
I can handle this issue.
Could you assign this issue to me?

PR #527, opened 2025-08-08
services/trustline - share button

pandablue0809:
# Issue
[Add a "share" button that automatically selects the current field settings in the browser link using query parameters.
](https://github.com/Bithomp/frontend-react/issues/493)

### DIFF
--- a/pages/services/trustline.js
+++ b/pages/services/trustline.js
@@ -374,11 +374,6 @@ export default function TrustSet({
       <SEO title="Set Trustline" description={'Set a Trustline on the ' + explorerName} />
       <div className="content-text content-center">
         <h1 className="center">Set/Update Trust (Trustlines)</h1>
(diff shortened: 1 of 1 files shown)


ihomp (changes requested):
1. Now when you open a page it has lots of default params added in the url
http://localhost:3000/en/services/trustline?currency=XRP&mode=simple&freeze=0&noRipple=1&authorized=0&deepFreeze=0

the default params shouldn't be shown in url, so the page should stay http://localhost:3000/en/services/trustline

check how it is done in the http://localhost:3000/en/nft-explorer—params are only added to the url, if they are different from the default. 

2. the button should be moved next to the "create trustline" button, when copied, the text shouldn't be added next to the button, but rather chnaged inside of the button .. 
"Share" -> changes to "Link copied" -> and then back to "Share"

ihomp (approved):

PR #529, opened 2025-08-08
xrpl-article updates

Anna15170221:
I tried to make it a bit more SEO friendly without changing the content and structure. Some paragraphs are a bit shortened. The article is replaced to learn section


pages/sitemap.xml.js
```diff
@@ -65,7 +65,8 @@ const pages = [
   { loc: 'learn/ripple-usd', changefreq: 'monthly', priority: '0.5' },
   { loc: 'learn/the-bithomp-explorer-advantages', changefreq: 'monthly', priority: '0.9' },
   { loc: 'learn/nft-minting', changefreq: 'monthly', priority: '1' },
-  { loc: 'learn/the-bithomp-api', changefreq: 'monthly', priority: '1' }
+  { loc: 'learn/the-bithomp-api', changefreq: 'monthly', priority: '1' },
+  { loc: 'learn/xrpl-article', changefreq: 'monthly', priority: '0.4' }
```
ihomp, line 69: No need to add here, as it will be added later only for xrpl.. 
if you add here, it will be included twice for xrpl and once for xahau, but we shouldn't have it on xahau

Issue #507, opened 2025-07-31
/services pages - display the filled parameters in url


Anna15170221:
/services pages 
We should display the filled parameters in url. 
Currently it works for /send page but we are missing this for: 
/check
/escrow
and some fields (like royalty) and flags on /nft-mint.

We should also do it for advanced options on /services pages and advanced mode on /trustline - but this is a lower priority and can be fulfilled later in a separate PR.

ihomp:
priority 4

PR #530, opened 2025-08-08
/services pages - display the filled parameters in url

pandablue0809:
# Issue
[We should display the filled parameters in url.
](https://github.com/Bithomp/frontend-react/issues/507)


ihomp (changes requested):
the default values shouldn't be added to the url, check how it is done in nft-explorer

components/Services/NftMint/NFTokenMint.js
```diff
@@ -69,9 +84,66 @@ export default function NFTokenMint({ setSignRequest, uriQuery, taxonQuery, acco
...
+      if (issuer) queryAddList.push({ name: 'issuer', value: issuer })
+      else queryRemoveList.push('issuer')
+    } else {
+      queryRemoveList.push('mintForOther')
+      queryRemoveList.push('issuer')
+    }
     addAndRemoveQueryParams(router, queryAddList, queryRemoveList)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [taxon, uri])
+  }, [taxon, uri, flags, transferFee, createSellOffer, amount, selectedToken?.currency, selectedToken?.issuer, destination, issuer, mintForOtherAccount, expiration])
```
ihomp, line 146: Couldn't we just specify selectedToken instead of selectedToken?.currency, selectedToken?.issuer ?

ihomp:
@Anna15170221 could you please test this branch? 

we ned to make sure nftokens add valid url params for expiration, for transfer fee, IOU token currency etc.. and then this url can be copied with those params and opened in a new window, and all params would be correctly filled.

ihomp (approved):

PR #532, opened 2025-08-11
XRP-XAH-TAXES article changed

Anna15170221:
1. The article is changed, some fresh info is added.
2. Replaced to learn section.
3. I made an attempt to add the redirect - not sure if it is ok.
4. New video is used for the article.


pages/sitemap.xml.js
```diff
@@ -84,7 +84,7 @@ if (network === 'mainnet') {
   if (!xahauNetwork) {
     pages.push({ loc: 'xrpl-article', changefreq: 'monthly', priority: '0.6' })
   }
-  pages.push({ loc: 'xrp-xah-taxes', changefreq: 'monthly', priority: '0.9' })
```
ihomp, line 87: @Anna15170221 here it was only in the sitemap for the mainnet (not pushed to sitemaps on testnets) which was good

Issue #488, opened 2025-07-24
/account page - sort dex orders by pair/size


Anna15170221:
/account page - sort dex orders by pair/size

ihomp:
we need to sort dex orders, similiar pairs should be next to each other..

if we have 
sell 1 BTH for 3.5 USD 
sell 1 BTH for 3.5 USD 
sell 1 BTC for 4000 USD, rate 1/4000
sell 1.2 BTC for 5000 USD, rate 1/4116
sell 1 BTH for 4 USD 

sort
1) by pair - similiar pairs together, next to each other
2) by rate - a lower rate for sell - top, a higher rate for buy - top
3) by size - larger one on top. 

so it will become 

sell 1 BTH for 3.5 USD 
sell 1 BTH for 3.5 USD 
sell 1 BTH for 4 USD 
sell 1.2 BTC for 5000 USD, rate 1/4116
sell 1 BTC for 4000 USD, rate 1/4000

ihomp:
https://bithomp.com/en/account/r3rhWeE31Jt5sWmi4QiGLMZnY3ENgqw96W

<img width="691" height="533" alt="Image" src="https://github.com/user-attachments/assets/ed75eeb2-d8ff-4116-a6d2-41e0dc008ab2" />

Let's change that:

- On the account page, lets just show "The last 5 DEX orders" with a [View all] link
- View All link should link to a new page /account/{address}/dex

On the /account page 
- we show only the last 5 orders (same as we had ordered by sequence, but only the latest 5)
- we need to show small token icons before to the currency name
- we can remove the service names and add them to tooltips (when hover on the token icon) - to save space
- we need to show shorter numbers, and add full number on the tooltip, right now it shows 0.7822540229885059 BTC - we need to show niceShortNumbers, but on hover show the full number. 
- We need to fix rates; on hover they should show the full numbers. Right now we have 1 FLR = 0 BTC, and we can't see the rate. 
- the rate when it brakes into two lines, doesn't brake nicely sometimes.. like
1 LTC = 110.76228
USD
we need to make sure that values stay next to the currency, like 
1 LTC =
110.76228 USD

On the new page: /account/{address}/dex
- we need to use a new endpoint, that checks only DEX orders objects (doesn't require all the objects) (Ask @ildaruz for it)
- we can show all the orders. 
- we can sort all the orders (by size, by time, by pair)
- we can filter all the orders (by currency, or pair, by time). 
- you can cancel any of them


priority 2.5

[...]

PR #535, opened 2025-08-13
/account page - sort dex orders by pair/size

pandablue0809:
# Issue
[/account page - sort dex orders by pair/size](https://github.com/Bithomp/frontend-react/issues/488)

### DIFF
--- a/components/Account/DexOrdersData.js
+++ b/components/Account/DexOrdersData.js
@@ -6,9 +6,48 @@ import { MdMoneyOff } from 'react-icons/md'
 export default function DexOrdersData({ account, offerList, ledgerTimestamp, setSignRequest }) {
   //show the section only if there are dex orders to show
   if (!offerList?.length) return ''
+  const isNativeAmount = (amt) => typeof amt === 'string'
+  const amountUnits = (amt) => (typeof amt === 'string' ? Number(amt) / 1000000 : Number(amt?.value || 0))
+  const amountCurrency = (amt) => (typeof amt === 'string' ? nativeCurrency : niceCurrency(amt?.currency))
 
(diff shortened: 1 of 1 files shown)


components/Account/DexOrdersData.js
```diff
@@ -1,28 +1,21 @@
...
+  // Sort offerList by sequence in ascending order and take only the last 5
+  const sortedOfferList = [...offerList].sort((a, b) => a.Sequence - b.Sequence).slice(-5)
 
   const historicalTitle = ledgerTimestamp ? (
     <span className="red bold"> Historical data ({fullDateAndTime(ledgerTimestamp)})</span>
   ) : (
     ''
   )
 
-  const offerListCountText = (offerList) => {
```
ihomp, line 19: it's nice to say how many offers were in general, like 

`The last 5 DEX offers [View all 35 offers]`
 or smth like that

ihomp:
1. do we have a timestamp or ledger number for orders? it would be nice to show timestamps instead of the sequence numbers. if not, its ok to have teh sequences. 
2. does it looks worse if the icon we put before the number in the formatamountIcon? 
3. we have text like 


The last 5 DEX orders [View all (2 total)] 

It is nice when there are more than 5 orders, but if orders  <=5, may be we should just show 

5 DEX orders [View details] 

4. right now there is no such page 
http://localhost:3000/en/account/r3rhWeE31Jt5sWmi4QiGLMZnY3ENgqw96W/dex

if we want to merge this PR, we should hide the link for now (while it is not ready), or we should create that page.

ihomp (changes requested):
check the comments

PR #539, opened 2025-08-16
fix token page UI

pandablue0809:


pandablue0809:
@ihomp 
I fixed token page
plz review again

Issue #520, opened 2025-08-05
Allow to open viewer in a separate window


Anna15170221:
Make possible to open the integrated viewer in a separate window (/nft page).

ihomp:
That is related to the 3d models, and to the images / video .. that can be separate PRs.

PR #548, opened 2025-08-19
Allow to open viewer in a separate window

pandablue0809:
# Issue
[Make possible to open the integrated viewer in a separate window (/nft page).](https://github.com/Bithomp/frontend-react/issues/520)

### DIFF
--- a/components/NftPreview.js
+++ b/components/NftPreview.js
@@ -7,9 +7,10 @@ import { needNftAgeCheck, nftName, nftUrl } from '../utils/nft'
 
 import Tabs from './Tabs'
 import LoadingGif from '../public/images/loading.gif'
-import { FaCloudDownloadAlt } from 'react-icons/fa'
+import { FaCloudDownloadAlt, FaExpand } from 'react-icons/fa'
 import ReactPannellum from 'react-pannellum'
 import AgeCheck from './UI/AgeCheck'
+import NftFullScreenViewer from './NftFullScreenViewer'
 
 const downloadIcon = (
   <div style={{ display: 'inline-block', verticalAlign: 'bottom', height: '19px' }}>
@@ -42,6 +43,7 @@ export default function NftPreview({ nft }) {
   const [errored, setErrored] = useState(false)
   const [isPanoramic, setIsPanoramic] = useState(false)
   const [showAgeCheck, setShowAgeCheck] = useState(false)
+  const [showFullScreen, setShowFullScreen] = useState(false)
 
   const style = {
     textAlign: 'center',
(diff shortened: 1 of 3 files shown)


ihomp:
@Anna15170221 can you test this one and write your suggestions? 

from me: it seems a lot of code was copy pasted from a different component, we shouldn't copy-paste the code, we need to extract it and reuse it.. that way we will keep the code maintanable. 

- the full-screen view - has trouble with scrolling; also, the scrolling button shouldn't be there. 
- the buttons to view full screen and to download music should be in the same style (new style you suggested) ..
- when we have both audio and image - we have two buttons to view full screen - that shoudn't be the case.

ihomp (changes requested):
It seems there is a lot of copy-pasted code that needs to be extracted and reused.

Anna15170221:
@pandablue0809 
Screenshot 2025-09-15 at 11 27 58
the buttons are on the image - for me doesn't look nice for desktop
and the mobile view should also be improved I think
Screenshot 2025-09-15 at 11 32 52

http://localhost:3000/en/nft/00083A9836EC5BC756007990B29DBFC81F45460ED8B8583A6442C30600000069 - I tested on this one

pandablue0809:
@ihomp 

pandablue0809:
@ihomp 
i fixed it
plz review again 🙏

components/NftPreview.js
```diff
@@ -262,92 +189,87 @@ export default function NftPreview({ nft }) {
...
-                <video autoPlay playsInline muted loop controls style={{ width: '100%', height: 'auto' }}>
+                <video autoPlay playsInline muted loop controls className="fv-preview-video">
                   <source src={videoUrl} type="video/mp4" />
                 </video>
               )}
             </>
           )}
           {modelUrl && defaultTab === 'model' && (
             <>
-              {modelState === 'loading' && (
```
ihomp: where is this part go?

ihomp:
There are many things affected by this PR: 
I see some logic was deleted for loading states for models for example.

We need to test: 
NFT pages with:
- image + audio 
- image + video 
- image + audio + video
- image + model
- model 
- video 

We need to make sure 
When the model is loading, we see the loading state on the NFT page 
When testing, we need to check how it is now and how it will be after the PR. 

we need to make sure we have proper loading states
- videos
- images 
- models 

We need to make sure we show errors for failing loading images 
that 18+ still works on the nft page 

@Anna15170221 @pandablue0809, please send me some examples where you've tested, so I can check too.

components/NftPreview.js
```diff
@@ -68,102 +56,17 @@ export default function NftPreview({ nft }) {
     }
   }
 
-  const imageUrl = nftUrl(nft, 'image')
-  const videoUrl = nftUrl(nft, 'video')
-  const audioUrl = nftUrl(nft, 'audio')
-  const modelUrl = nftUrl(nft, 'model')
-  const viewerUrl = nftUrl(nft, 'viewer')
+  const urls = extractNftUrls(nft)
+  const { image: imageUrl, video: videoUrl, audio: audioUrl, model: modelUrl, viewer: viewerUrl } = urls
```
ihomp: why , name it right away as it will be used.
Instead of reassigning teh variables, just update teh function

Issue #519, opened 2025-08-05
Add option to send tokens with Remit


Anna15170221:
To add an option to send tokens with remit on xahau (then any token can be sent if destination has incoming remit not disabled, sender pays for destination reserves).

ihomp:
That is for the page /send, to send IOU.

PR #549, opened 2025-08-19
Add option to send tokens with Remit

pandablue0809:
# issue
[To add an option to send tokens with remit on xahau (then any token can be sent if destination has incoming remit not disabled, sender pays for destination reserves).](https://github.com/Bithomp/frontend-react/issues/519)

### DIFF
--- a/pages/services/send.js
+++ b/pages/services/send.js
@@ -427,14 +427,7 @@ export default function Send({
 
   return (
     <>
-      <SEO
-        title="Send payment"
-        description={
-          xahauNetwork
-            ? 'Send a payment to a destination address. On Xahau network, use Remit to send any token (including native XAH) to destinations with incoming remit enabled.'
-            : 'Send a payment to a destination address'
-        }
-      />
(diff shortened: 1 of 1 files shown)


ihomp:
@Anna15170221 do you have an example of the successful transaction from yesterday?

ihomp:
tested

- it's might be we are connected to an old node, that can't send without pathfinding... (I've asked @ildaruz to check it)

so we will test it later again, when he answers. 

1) lets remove network switch tabs 
2) lets move Use Remit before the token selection 
3) chnage the text: "- Send any token to destinations with incoming Remit enabled. Sender pays for destination reserves." to "- Send any token and pay for destination reserves." (if remit is not enabled, it will be disabled and show the message here anyway)
4) remove the text on top: 💡 Xahau Network: Use the Remit option below to send any token (including native XAH) to destinations with incoming remit enabled - we already have this text next to remit checkbox

ihomp (approved):

ihomp (approved):

ihomp (approved):

Issue #518, opened 2025-08-05
Add NFT Transfer Option on Xahau Network using the Remit.


Anna15170221:
We need to add an option, to transfer NFT using the Remit. If destination doesn't have disallowRemitIncoming (allows remit trxs) then we can just send NFT to the destination without asking (but paying for the NFT reserve).

ihomp:
@Anna15170221 don't we already have that on xahau? 

What we need is a new option to transfer using Rimit (when the user pays for xah reserve on the destination account for that nft). 

We should have an option now: to transfer. 1) Create a sell offer with a price 0 and a destination. 2) then the destination buys it (accepts the offer)

We need to add an extra option, to transfer using the Remit.. if destination doesn't have disallowRemitIncoming (allows remit trxs) then we can just send NFT to the destination without asking (but paying for the NFT reserve).

PR #550, opened 2025-08-20
Add NFT Transfer Option on Xahau Network using the Remit. 

pandablue0809:
# Issue
[We need to add an option, to transfer NFT using the Remit. If destination doesn't have disallowRemitIncoming (allows remit trxs) then we can just send NFT to the destination without asking (but paying for the NFT reserve).](https://github.com/Bithomp/frontend-react/issues/518)

### DIFF
--- a/components/SignForm.js
+++ b/components/SignForm.js
@@ -171,15 +171,30 @@ export default function SignForm({
     }
 
     if (signRequest.action === 'nftTransfer') {
-      tx.Amount = '0'
-      if (!agreedToRisks) {
-        setScreen('nftTransfer')
-        return
-      } else {
-        if (!signRequest.request?.Destination) {
-          setStatus(t('form.error.address-empty'))
-          setFormError(true)
+      if (tx.TransactionType === 'Remit') {
+        // For Remit transactions, no Amount field is needed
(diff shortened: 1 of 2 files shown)


components/SignForm.js
```diff
@@ -171,15 +171,30 @@ export default function SignForm({
...
+            setFormError(true)
+            return
+          }
+        }
       } else {
-        if (!signRequest.request?.Destination) {
-          setStatus(t('form.error.address-empty'))
-          setFormError(true)
+        // For regular NFT transfer offers
+        tx.Amount = '0'
```
ihomp, line 188: This is code is exactly the same as for remit, except the amount.. Please don't copy-paste the same code!

```
if (tx.TransactionType !== 'Remit') {
  tx.Amount = '0'
}
```

ihomp:
@Anna15170221 please test this one too

ihomp (changes requested):
requires a bit of code optimisation

Anna15170221:
@ihomp @pandablue0809 
seems not working for me
Tx is signed and I got this
Screenshot 2025-09-15 at 13 16 18

pandablue0809:
@ihomp 

Issue #559, opened 2025-08-22
/tokens sorting options improvements needed


Anna15170221:
instead of sorting icons as two side arrows, we can show only one arrow if we can sort down (arrow down), orange when active, if not sorted, but possible two sides - two arrows green - when sorted arrow up or down - depends which way is sorted by.

PR #571, opened 2025-08-26
when you choose sorting in the dropdown, the arrows need to be updated.

pandablue0809:
# Issue
https://github.com/Bithomp/frontend-react/issues/559#issuecomment-3224531374


ihomp (commented):
what is this? that seems just reverts the merged changes.. 

why do you want to revert the merged PR?

Issue #517, opened 2025-08-04
/tokens page - improvements needed


Anna15170221:
for the /tokens
Separate PRs (all priority 1): 
1. We have a mess with the param `const [filtersHide, setFiltersHide]` 
The problem that those two behaves differently and means the opposite on mobile. On mobile right now `filtersHide ` means filters will be open. We need to fix it, so filtersHide as true should mean that filters are hidden - doesn’t matter if it’s desktop or mobile. For now it was just added a quick hack and set by default the value as `!isSsrMobile const [filtersHide, setFiltersHide] = useState(!isSsrMobile)`   which is ok only on mobiles and destktops - but very messy on other screens - the filters will be open automatically on the page load. so, that param should be fixed.. `filtersHide` should mean no filters on both mobile and desktop. As it was extracted as a component - we would need to make the changes in all the places where it is in use in the app. 
2. we now have the prices in xrp for each token, 5minutes ago, 1 hour ago etc.. for example we have price of ATM token in XRP 5m ago.
we need to receive historical XRP price for 5 min ago in fiatCurrency that user use. (we use historical endpoints in the react already - if you cant find ask Ildar for it ) that way you can calculate how much in fiat (USD/EUR/GBP,JPY etc) the token was worth 5min ago, 1h ago etc.. Then we can calculate how much token grow or fall since then. current fiat price / historical price of the token in fiat. we need to show this percentage numbers in the table.
3. we need to make all columns sortable (instead of the sorting options on top right) similar to how we have it in the http://bithomp.com/en/nft-volumes?period=week (Ildar will add the unavailable yet options, for now we can do the ones that are available.) the rating sorting is for the index column.

ihomp:
1. Instead of fixing the hack from the previous consultant, you've just added an extra hack =) 

```
          @media (max-width: 1300px) {
            /* Invert mobile filter show/hide semantics for Tokens page only */
            .tokens-page .content-cols.is-filters-hide .filters__wrap {
              transform: translateY(100%) !important; /* hide when filtersHide === true */
            }
            .tokens-page .content-cols:not(.is-filters-hide) .filters__wrap {
              transform: translateY(0) !important; /* show when filtersHide === false */
            }
          }
```

1) This doesn't work omn screens over 860/1080 px - that was set inside the page for different table views 
2) That BRAKES the infinite scroll, so you wont be able to see more than 100 tokens 

So, I've removed it.

ihomp:
The same code was added twice 

```
[...]

PR #573, opened 2025-08-26
Fix/517/update filter hide

pandablue0809:
# Issue
[We have a mess with the param const [filtersHide, setFiltersHide]
The problem that those two behaves differently and means the opposite on mobile. On mobile right now filtersHide  means filters will be open. We need to fix it, so filtersHide as true should mean that filters are hidden - doesn’t matter if it’s desktop or mobile. For now it was just added a quick hack and set by default the value as !isSsrMobile const [filtersHide, setFiltersHide] = useState(!isSsrMobile) which is ok only on mobiles and destktops - but very messy on other screens - the filters will be open automatically on the page load. so, that param should be fixed.. filtersHide should mean no filters on both mobile and desktop. As it was extracted as a component - we would need to make the changes in all the places where it is in use in the app.](https://github.com/Bithomp/frontend-react/issues/517)


ihomp:
that is nt what we need, it is just changing the true/false state and make it oposite.. 

the problem that `filtersHide` as `true` should hide the filters on both desktop and mobile. 
filtersHide as false should open filters on both desktop and mobile 

we should keep, that on small screens its hidden, and open on desktops

Issue #476, opened 2025-07-23
Issue your own currency - new page


Anna15170221:
We need to create a new page to issue your own currency with all the necessary buttons.
We need to write the text to explain how to do it.
This page will be available to Bithomp Pro subscribers only.

PR #593, opened 2025-09-07
Issue your own currency - new page

pandablue0809:
# Issue
[create a new page to issue your own currency](https://github.com/Bithomp/frontend-react/issues/476)


components/Layout/Header/index.js
```diff
@@ -196,11 +196,13 @@ export default function Header({
             <Link href="/services/send">Send Payment</Link>
             <Link href="/services/trustline">Set Trust (Trustline)</Link>
             <Link href="/services/check">Issue Check</Link>
+            <Link href="/services/issue-currency">Issue Currency</Link>
             <Link href="/services/escrow">Create Escrow</Link>
             {!xahauNetwork && !devNet && <Link href="/services/amm">AMM Services</Link>}
             <Link href="/services/account-settings/">Account Settings</Link>
             <Link href="/services/nft-mint">{t('menu.services.nft-mint')}</Link>
             <Link href="/username">{t('menu.usernames')}</Link>
+            <Link href="/xrp-xah-taxes">{t('menu.services.tax-reports')}</Link>
```
ihomp, line 205: should it be in the same PR?

ihomp:
@Anna15170221 can you help testing this one and give your feedback

ihomp:
@Anna15170221 can you please test again, thx

Anna15170221:
@pandablue0809 @ihomp 
- The "Configure Cold/HOT Wallet" button is always inactive. Should it be like this?

https://github.com/user-attachments/assets/7c82b8f7-313c-4a6c-b5e5-e19665586108

- if I press back to previous step, it erases some fields on the previous page (the address for example)

- We have Transfer rate in both Cold and Hot Wallet configurations - should it be like this? we shouldn't mention it in Hot Wallet Configs, I guess. https://xrpl.org/docs/tutorials/how-tos/use-tokens/issue-a-fungible-token. And it can be 0-1%, so the user can't be allowed to enter something different. @ihomp Am i right? 

- I think we are missing the explanation for settings, for tick size and others.

- If on the previous step we specified that the limit can be 1000, why do we mention that it can be 1000 or higher? 
Screenshot 2025-09-30 at 17 20 55

- Step 3: I think we need to make it possible to create a trustline right away on this page. Otherwise it's a gap in the whole process and the user will need to go somewhere else to create the trustline. And after setting the trustline, the user will send the token from one wallet to another. - He also should have an opportunity to do it right away on this page as the part of the process. 
At this moment Step 3 seems to be not finalised at all and the buttons are inactive. 
Screenshot 2025-09-30 at 17 27 59

pandablue0809:
@Anna15170221 @ihomp 
i fixed it.
and if i input the wallet address, the "Configure Cold/Host Wallet" is active.
plz review again. 🙏

Issue #579, opened 2025-08-28
balance change history - open data for several wallets


Anna15170221:
Currently this page (https://bithomp.com/en/admin/pro/history) is available for PRO subscribers only. 
We need to add there several wallets and make the page open for logged out users. 
On this page we should mention that adding user's own wallets is available for logged-in Bithomp Pro subscribers. And give the link to sign in to PRO. After login the user should be redirected to the page where he can add his wallets (https://bithomp.com/en/admin/pro)

The wallets which are currently available for non-authorised users: 
https://bithomp.com/explorer/rDLNm4ehD7XQCtYKWuMjEKY7TCfmf3CwzH
https://bithomp.com/explorer/r3sQYvXxc82iSNs5DnUUvXtcQQQqigCdW
https://bithomp.com/explorer/rhWTXC2m2gGGA9WozUaoMm6kLAVPb1tcS3
https://bithomp.com/explorer/raWYT6DD2XFAvjCqRPsCCzr1CMBzJydf9E
https://bithomp.com/explorer/rakZprdzwsUJ1rD2ouhYYAVP7tPbhrCbtz
https://bithomp.com/explorer/r3LAichpcBeZWk7LLSZcfSQcqYsvQ6beBc
Ildar has already added them. 

On the screenshot you see 2 added addresses in the left part. The open page will show 6 addresses from the list above (probably we will make the list less after the testing) and the user should have an opportunity to try all the functions. To test on 1 or several (or all) addresses simultaneously. To download all available reports, test all available filters. 

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/dd419634-28bb-4c27-b9dc-0feca12600a1" />

ihomp:
@Anna15170221, closing this, if any issues - dm to me directly

PR #607, opened 2025-09-11
balance change history - open data for several wallets

pandablue0809:
# Issue
[add several wallets in balance change history.](https://github.com/Bithomp/frontend-react/issues/579)


pandablue0809:
@ihomp 

ihomp:
Screenshot 2025-09-26 at 09 34 21

There seem like lots of troubles. 
- LP tokens don't show the pairs. 
- LP tokens don't always show the value, says NAN 
- the sorting by largest FIAT or largest XRP values seems broken too
- when changing sorting - loaded transactions are erased, but no notification about it, easy to miss. if not all transactions are loaded - then sorting is invalid - may be we should diasble sorting when not all transactions loaded 
- there are more than a million transactions, I guess 10k would be enough for all the wallets. may be we can remove some wallets 

@Anna15170221 

In general, we can move all this functionality to a new website right away.

pandablue0809:
@ihomp 

ihomp (changes requested):
resolve conflicts

pandablue0809:
@ihomp 
i fixed it. plz review again 🙏

Issue #602, opened 2025-09-09
/token page - if the price is less than 1 drop


Anna15170221:
example: https://bithomp.com/en/token/rJMtvf5B3GbuFMrqybh5wYVXEH4QE8VyU1/586F676500000000000000000000000000000000
1. If the price is less than 1 drop, multiply by 1,000,000 and show 1M XOGE = X XRP

2. Also display reverse prices: 1 XRP = X XOGE

PR #609, opened 2025-09-11
/token page - if the price is less than 1 drop

pandablue0809:
# Issue
[update the price if the it is less than 1 drop.](https://github.com/Bithomp/frontend-react/issues/602)


ihomp:
@Anna15170221 pls give your feedback on this one too

Anna15170221:
@ihomp @pandablue0809 
we will need to check how it looks on mobile. I locally see this but probably it will look different on mobile
Screenshot 2025-09-15 at 15 08 06

ihomp (approved):

Issue #556, opened 2025-08-21
/transactions page - fixes needed


Anna15170221:
- no limitation for infinite scroll for non-pro accounts (right now we show the message that they cant see more than 20 transactions - that shouldn’t be the case)
- the sorting newest first /latest first doesn't work
- csv export doesn’t work
- divide each separate transaction by the line. how it's made in php. Otherwise it's hard to read.
- leave the phrase Transaction hash: near the hash. It's absolutely needed for newcomers.
- We used in php two arrows symbol only for exchanges. Currently it's used for all transactions. We have different icons in balance change history - let's reuse them for transactions.
- Failed transactions are almost unreadable (orange on orange) and for the dark theme it's not visible at all.
- For NFToken Burn txs we showed: Change: removed NFT and then NFT ID. please keep it like it was before. Currently we have only NFT ID.
- For NFTokenBurn no need to show flags like XRPOnly, transferrable and so on. In php we showed the flags for NFTokenMint but not for NFTokenBurn.
- Don't call Royalty Transfer fee. Call it Royalty.  
- Don't show source tag for every tx, show it only where it was mentioned in tx.
- show sequence (we have it in php for every tx)
- We showed small icons for tokens if they had one in php, let's keep it.

Transactions were tested on this account https://bithomp.com/en/account/rfjquSFSKXnDYdQyGWuXBSSSW87jfpX7Qm/transactions and I provided hashes where I found the bugs or missing info.
- https://bithomp.com/en/tx/644318418059A55B5B84B0067F0C44F56C0E5EA256FA34BA8DF6BF356AB67DE6 If destination tag is 0 like here, on /transactions it shows just 0, without mentioning that it's the destination tag. Looks weird. Screen attached.

<img width="755" height="135" alt="Image" src="https://github.com/user-attachments/assets/9881f737-276f-432e-a967-4684f3e74d86" />

- NFTokenCreateOffer - Sell Offer (if the offer was not initiated by this account, we should show who was the offer creator, like in php). NFTokenCreateOffer - Sell Offer by .....-username or address ([6EAE982A488A370BDF5784216E00F6FC2A246516B079C9158637040612EC1B6A](https://bithomp.com/explorer/6EAE982A488A370BDF5784216E00F6FC2A246516B079C9158637040612EC1B6A))
[...]

PR #615, opened 2025-09-16
transactions page - fixes needed 

pandablue0809:
# Issue
[1. In TrustSet we don't show the token, the limit, the flags like we do in php.
2. In AccountSet we don't show what changed - which flags were set/removed/updated
3. I would suggest to put icon first and then put "+" to make it look similar with "-" (see the screenshot)
4. small issue: failed txs look readable now, but still a bit hard to read the orange, probably we could play with colours a bit. But on mobile looks nice at the moment.
5. Don't show all flags and royalty in NFTokenBurn transactions.
6. I would add the crossmark to the date/time fields to reset the filled data faster.
7. Ideally If the user sets some filters, then opens the transaction page/nft page/nft offer page and then returns back - i would keep the same filters adjusted. so that he won't need to filter again and just continue the searching.
8. the source tag is shown with 0 in the begining
https://bithomp.com/en/account/MainForTesting/transactions
9. TicketSequence is displayed incorrectly in the transaction list.
https://bithomp.com/en/account/rpiFwLYi6Gb1ESHYorn2QG1WU5vw2u4exQ/transactions](https://github.com/Bithomp/frontend-react/issues/556)

### DIFF
--- a/components/Transactions/TransactionRowOffer.js
+++ b/components/Transactions/TransactionRowOffer.js
@@ -4,13 +4,12 @@ import { addressBalanceChanges } from '../../utils/transaction'
 import { amountFormat, nativeCurrencyToFiat, addressUsernameOrServiceLink } from '../../utils/format'
 
 const flagList = (flags) => {
-    
   let flagsString = ''
 
   if (!flags) return flagsString
 
   for (let key in flags) {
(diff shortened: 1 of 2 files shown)


pandablue0809:
@ihomp 

Issue #587, opened 2025-09-05
Batch transactions - Add support


Anna15170221:
Add support of batch txs 
https://xrpl.org/es-es/docs/references/protocol/transactions/types/batch
example
https://dev.bithomp.com/en/tx/0C5956CF40BD544278D72697FF3FAC00571A52196252E006A2EDB8381D96E252 
```

/v3/transaction/:hash

support for batch txs:
specification.transactions[]

each transaction will have formatted specification data
transaction: 
  id - generated on our side from RawTransaction data
  address - from RawTransaction
  sequence - from RawTransaction
  type - from RawTransaction
  specification - generated on our side from RawTransaction data

specification.flags -- all transactions will have its own type depended flags, including Global ones
Note: global flags are diffrrent for XRP/XAH networks
example: https://dev.bithomp.com/api/cors/v3/transaction/0C5956CF40BD544278D72697FF3FAC00571A52196252E006A2EDB8381D96E252
```

```
/v3/transaction/:hash

tx creadted by batch will have hash from original transactions
outcome.parentBatchID
example: https://dev.bithomp.com/api/cors/v3/transaction/C8D02BA16B50C5BD12126F6864EB14AF1A33B31DE9053CAC5C5667BD77A37719
```
The difference with "normal" transactions: it will have a new key in specification with batch basic transactions data
To recognise if its a batch: flags, and transaction flags
The support on devnet is needed. 
Example:
https://github.com/Bithomp/xrpl-api/blob/3851db65bd6c9c2767e7b54a7ec722171df117e6/test/examples/responses/Batch.json
To find more examples:
https://github.com/Bithomp/xrpl-api/blob/master/test/models/transaction.test.ts#L7033

PR #621, opened 2025-09-19
Batch transactions - Add support

pandablue0809:
# Issue
[Add support of batch txs](https://github.com/Bithomp/frontend-react/issues/587)
Now, I display the main datas.
If you share your specific requirements, I’ll take care of the implementation.

### DIFF
--- a/components/Transaction/TransactionBatch.js
+++ b/components/Transaction/TransactionBatch.js
@@ -1,0 +1,162 @@
+import React from 'react'
+import { TData } from '../Table'
+
+import { TransactionCard } from './TransactionCard'
+import { addressUsernameOrServiceLink, AddressWithIconFilled } from '../../utils/format'
+import { xahauNetwork } from '../../utils'
+
+// Global flags definitions for XRP and XAH networks
+const GLOBAL_FLAGS_XRP = {
+  tfFullyCanonicalSig: 'Require fully canonical signatures',
+  tfNoDirectRipple: 'Pathfinding is not used to find this path',
+  tfPartialPayment: 'Partial payment is allowed',
+  tfLimitQuality: 'Limit quality is used for this offer',
+  tfFillOrKill: 'Fill or kill order',
+  tfImmediateOrCancel: 'Immediate or cancel order',
+  tfSell: 'Sell order',
+  tfPassive: 'Passive order',
+  tfRequireDestTag: 'Require destination tag',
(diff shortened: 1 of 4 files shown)


Anna15170221:
@ihomp 
should we probably show fiat amounts also in details?

ihomp:
> @ihomp should we probably show fiat amounts also in details?

Yes, but I think it will only be visible in production (mainnet/staging) - we can test with "staging" instead of "devnet" - than it should show.

pandablue0809:
@ihomp 
i fixed it.
plz review again. 🙏

components/Transaction/TransactionBatch.js
```diff
@@ -0,0 +1,93 @@
...
+    <TransactionCard
+      data={data}
+      pageFiatRate={pageFiatRate}
+      selectedCurrency={selectedCurrency}
+      notFullySupported={true}
+    >
+      <tr>
+        <TData>Initiated by</TData>
+        <TData>
+          <AddressWithIconFilled data={specification.source} name="address" />
```
ihomp: that seems used wrong 
it should be object and then the name of the address field 

`<AddressWithIconFilled data={specification} name="source" />`

pandablue0809:
@ihomp 
I fixed it.
plz review again. 🙏

components/Transaction/TransactionBatch.js
```diff
@@ -0,0 +1,97 @@
...
+        </TData>
+      </tr>
+      {outcome?.parentBatchID && (
+        <tr>
+          <TData>Parent batch ID</TData>
+          <TData>{addressUsernameOrServiceLink(outcome, 'parentBatchID', { short: true, url: '/tx/' })}</TData>
```
ihomp: Why do you use here `addressUsernameOrServiceLink` that is supposed to be used for addresses, not for transactions? 

for transactions you can use LinkTx

components/Transaction/TransactionBatch.js
```diff
@@ -0,0 +1,97 @@
...
+            </TData>
+          </tr>
+          <tr>
+            <TData>Flags</TData>
+            <TData>
+              {showFlags(transaction?.specification?.flags)}
+            </TData>
+          </tr>
+          {
+            transaction.specification.destination &&
```
ihomp: shouldn't it be transaction?.specification?.destination as you have it that way in other places 

if specification doesn't exist it will brake js

Issue #560, opened 2025-08-22
we need to show MPToken properly


Anna15170221:
https://dev.bithomp.com/en/account/rEpeYYLHGGAiQhs3zKRjmyfMQJg9HyvYgc
This can be used to fix the page:

`mptokenCurrencyDetails for mptoken object`

```

https://dev.bithomp.com/en/object/A207C1A3C16F60797A065FF74EA472141B0D6F7F9091B2825834CD2DFFE1EEF6

/v2/ledgerEntry/A207C1A3C16F60797A065FF74EA472141B0D6F7F9091B2825834CD2DFFE1EEF6?currencyDetails=true
```

example

    "mptokenCurrencyDetails": {
      "type": "mp_token",
      "mptokenIssuanceID": "0042AB9EAB8A5036CE4DB80D47016F557F9BFC9523985BF1",
      "account": "rGepNyxjJbtN75Zb4fgkjQsnv3UUcbp45E",
      "accountDetails": {
        "address": "rGepNyxjJbtN75Zb4fgkjQsnv3UUcbp45E",
        "username": null,
        "service": null
      },
      "scale": 2,
      "currency": "RLUSD",
      "metadata": {
        "currency": "RLUSD",
        "name": "Ripple USD",
        "desc": "A regulated stablecoin issued by Ripple.",
        "icon": "https://ripple.com/assets/rlusd-logo.png",
        "asset_class": "rwa",
        "asset_subclass": "stablecoin",
        "acct_name": "Ripple",
        "weblinks": [
          {
            "url": "https://ripple.com/",
            "type": "website",
            "title": "Official Website"
          },
          {
            "url": "https://xrpl.org/",
            "type": "docs",
            "title": "XRPL Documentation"
          }
        ]
      }
    },

keep in mind, in this object example metadata is not following specification
https://github.com/XRPLF/XRPL-Standards/discussions/264 

for mpt escrow works too

 ```
Amount: {
              mpt_issuance_id: "0042AB9EAB8A5036CE4DB80D47016F557F9BFC9523985BF1",
              value: "589",
              currencyDetails: {
                type: "mp_token",
                mptokenIssuanceID: "0042AB9EAB8A5036CE4DB80D47016F557F9BFC9523985BF1",
                account: "rGepNyxjJbtN75Zb4fgkjQsnv3UUcbp45E",
                accountDetails: { address: "rGepNyxjJbtN75Zb4fgkjQsnv3UUcbp45E", username: null, service: null },
[...]

PR #623, opened 2025-09-23
we need to show MPToken properly

pandablue0809:
# Issue
[fix showing MPToken](https://github.com/Bithomp/frontend-react/issues/560)

### DIFF
--- a/components/Account/ObjectsData.js
+++ b/components/Account/ObjectsData.js
@@ -101,9 +101,8 @@ export default function ObjectsData({
         .get(
           'v2/objects/' +
             address +
-            '?limit=1000&priceNativeCurrencySpot=true' +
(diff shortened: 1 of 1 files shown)


ihomp:
we don't need to send an extra request for each mpt escrow, you need to send `?currencyDetails=true` directly to the endpoint, where we receive objects, so all the objects will have currencyDetails included 
: 'v2/ledgerEntry/' + escrow.index + '?currencyDetails=true' `

if the flag doesn't work for you for objects, ask @ildaruz to check

ihomp:
Screenshot 2025-10-02 at 17 18 13

That looks broken because of the MPT token IDs. 

what we had to chage here, it's instead of MPTIDs in the amounts, to show amounts as we for other escrows 
like 589 USDRL with the issuer like we do for iou escrows.

Issue #551, opened 2025-08-20
NFT Collection page


Anna15170221:
We need to create a separate page for each NFT Collection.
https://bithomp.com/en/nft-volumes?period=week - this is the page where we show the list of all collections. 
By clicking on collection link, the user will see a new page  with Collection picture, name, description, Stats and all collection items.
So it could look similar to /nft page but instead of the history it will show the collection items. 
I would also include a mini-activity feed for 3 recent sales/ 3 recent listings/ 3 recent mints in the collection. @ihomp do you agree?

PR #624, opened 2025-09-24
NFT Collection page

pandablue0809:
# Issue
[Create NFT collection page](https://github.com/Bithomp/frontend-react/issues/551)

### DIFF
--- a/pages/nft-collection/[id].js
+++ b/pages/nft-collection/[id].js
@@ -2154,7 +2154,7 @@ export default function NftCollection({ id, nftList, selectedCurrency, isSsrMobi
                     <tr>
                       <td style={{ width: 70 }}>NFT</td>
                       <td>
-                        <NftImage nft={item} style={{ width: 20, height: 20, objectFit: 'cover', borderRadius: 4 }} />{' '}
+                        <NftImage nft={item} style={{ width: 20, height: 20, borderRadius: 4 }} />
                         <Link href={'/nft/' + item.nftokenID}>{nftName(item?.nftoken || item)}</Link>
                       </td>
                     </tr>
@@ -2273,16 +2273,11 @@ export default function NftCollection({ id, nftList, selectedCurrency, isSsrMobi
           {items.length > 0 ? (
             items.map((item, i) => (
               <tr key={i}>
-                <td>
-                  <div style={{ display: 'flex', alignItems: 'center', gap: '10px', width: '100%' }}>
-                    <NftImage
-                      nft={item}
-                      style={{ width: nftImageSize, height: nftImageSize, objectFit: 'cover', borderRadius: 4 }}
-                    />
-                    <span className="code" style={{ display: 'flex', flexDirection: 'column' }}>
-                      <Link href={'/nft/' + item.nftokenID}>{nftName(item?.nftoken || item)}</Link>
-                    </span>
-                  </div>
+                <td style={{ display: 'flex', alignItems: 'center' }}>
+                  <NftImage nft={item} style={{ width: nftImageSize, height: nftImageSize, borderRadius: 4 }} />
+                  <span style={{ verticalAlign: 'middle' }}>
+                    <Link href={'/nft/' + item.nftokenID}>{nftName(item?.nftoken || item)}</Link>
+                  </span>
                 </td>
                 {kind === 'sales' && (
                   <>
--- a/utils/nft.js
+++ b/utils/nft.js
@@ -529,11 +529,14 @@ export const NftImage = ({ nft, style }) => {
      </text>
    </svg>`
   )}`
+  if (style?.width === 20) {
+    style.marginBottom = '-5px'
+  }
   return (
     <img
-      src={nftUrl(nft?.nftoken || nft, 'thumbnail') || placeholder}
+      src={nftUrl(nft?.nftoken || nft, size < 32 ? 'thumbnail' : 'preview') || placeholder}
       alt={nftName(nft?.nftoken || nft) || 'NFT thumbnail'}
-      style={style ? style : null}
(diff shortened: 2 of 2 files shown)


ihomp:
All the nft-collections page I open says: no image available, lots of empty space on the left side.. 

on the top I see some nfts from collections, but there is no link to view all of them

ihomp:
Agree with all the points @Anna15170221 ! 
Please update @pandablue0809

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,619 @@
...
+    setActivityLoading(true)
+
+    try {
+      const [salesRes, listingsRes, mintsRes] = await Promise.all([
+        axios(
+          `/v2/nft-sales?collection=${encodeURIComponent(id)}&list=lastSold&limit=3&convertCurrencies=${selectedCurrency}`
+        ).catch(() => null),
+    
+        axios(
+          `/v2/nfts?collection=${encodeURIComponent(id)}&list=onSale&order=offerCreatedNew&limit=3&currency=${selectedCurrency}`
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: that endpoint was already called on th server side, just get data from there.. you can use `serverCurrency` function from our utils to get the selectedCurrency from cookie. we have it done on other pages.

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,619 @@
...
+    return (
+      data?.collection?.name || (
+        <>
+          {addressUsernameOrServiceLink(data?.collection, 'issuer', { short: isMobile })} ({data?.collection?.taxon})
+        </>
+      )
+    )
+  }
+
+  const collectionDescription = (data) => {
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: The `data` is constant; there is no need to do a function 

`const collectionDescription = data?.collection?.description || ''`

ihomp:
1) Check the comments in the code

2) on the server side you are calling two unrelated api endpoints that are wait for each other, they shouldn't wait for each other, as they are not depended one on another, you need to call them async to safe time. 

3) check how nfts previews look on the /account page on desktop and mobile, they should look the same here. 
the size should be the same, so it will ne the same space on the left and right, it should look good on both desktop and mobile

pandablue0809:
@ihomp 
i fixed it.
plz review again. 🙏

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,610 @@
...
+    ])
+  } catch(error) {
+    errorMessage = "error." + error.message
+  }
+
+  if (!dataRes?.data) errorMessage = 'No data found'
+
+  return {
+    props: {
+      id: collectionId,
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: it's better to specify null here, not ''.. 

so it will be 

id: collectionId || null,
data: dataRes?.data || null,
errorMessage: errorMessage || null,

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,610 @@
...
+  const collection = data?.collection
+  const [activityData, setActivityData] = useState({
+    sales: [],
+    listings: [],
+    mints: []
+  })
+  const [activityLoading, setActivityLoading] = useState(false)
+  const [isMobile, setIsMobile] = useState(isSsrMobile)
+
+  useEffect(() => {
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: That is a good thing, but in general, it should be extracted as a function so it can be reused on other pages. 

it's ok to leave it for now

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,610 @@
...
+                                {amountFormat(item?.sellOffers?.[0]?.amount)}
+                                {nativeCurrencyToFiat({
+                                  amount: item?.sellOffers?.[0]?.amount,
+                                  selectedCurrency,
+                                  fiatRate
+                                })}
+                              </div>
+                              <div style={{ display: 'flex', alignItems: 'center', gap: '4px' }}>
+                                <span>Date: </span>
+                                {new Date(item.ownerChangedAt * 1000).toLocaleDateString()}
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: we have date functions in utils.js or format.js, they already use toLocaleDateString() - so please reuse them to keep the app look similiar and maintanable

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,610 @@
...
+                        <AddressWithIconFilled data={item} name="seller" options={{ short: true }} />
+                      ) : (
+                        '—'
+                      )}
+                      {item.buyer ? <AddressWithIconFilled data={item} name="buyer" options={{ short: true }} /> : '—'}
+                    </td>
+                    <td>
+                      {amountFormat(item.amount)}≈ {convertedAmount(item, selectedCurrency, { short: true })}
+                    </td>
+                    <td>{item.acceptedAt ? new Date(item.acceptedAt * 1000).toLocaleDateString() : 'N/A'}</td>
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: we have date functions in utils.js or format.js, they already use toLocaleDateString() - so please reuse them to keep the app look similiar and maintanable

pages/nft-collection/[id].js
```diff
@@ -0,0 +1,610 @@
...
+
+                        <div className="column-right">
+                          <div className="collection-info">
+                            <h2 className="collection-name">{collectionName(data)}</h2>
+                            {collectionDescription && (
+                              <p className="collection-description">{collectionDescription}</p>
+                            )}
+                          </div>
+
+                          {mounted && data?.collection?.issuerDetails && (
```
Comment (ihomp) [inline pages/nft-collection/[id].js]: you have collection = data?.collection

so you can use collection?.issuerDetails

don't check for issuerDetails, check for issuer, as we not nessary have issuerDetails, but we might still have issuer

ihomp (changes requested):
Please check the comments

Issue #538, opened 2025-08-15
the search by paystring and X-address is broken in React


Anna15170221:
the search by paystring and X-address is broken in React (works in php) - this needs to be fixed
https://bithomp.com/explorer/username$paystring.crypto.com
https://bithomp.com/explorer/XVVFXHFdehYhofb7XRWeJYV6kjTEwbq2mLScCiYyDTHKu9E

PR #625, opened 2025-09-29
the search by paystring and X-address is broken in React

pandablue0809:
# Issue
[1. Resolve paystring / xAddress (you need to find all the details, similiar to what you get in searchBlock), the most important part is the TAG.
2. if xAddress/paystring has a TAG, then show a new VIEW similiar to php
3. if no tag, show normal account page](https://github.com/Bithomp/frontend-react/issues/538)

### DIFF
--- a/pages/account/[id].js
+++ b/pages/account/[id].js
@@ -6,7 +6,16 @@ import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import Image from 'next/image'
 import { axiosServer, getFiatRateServer, passHeaders } from '../../utils/axios'
 
-import { devNet, xahauNetwork, avatarSrc, nativeCurrency, errorT, isValidPayString, isValidXAddress } from '../../utils'
+import {
+  devNet,
+  xahauNetwork,
+  avatarSrc,
+  nativeCurrency,
+  errorT,
+  isValidPayString,
+  isValidXAddress,
+  isTagValid
+} from '../../utils'
 import { shortNiceNumber } from '../../utils/format'
 import { getIsSsrMobile } from '../../utils/mobile'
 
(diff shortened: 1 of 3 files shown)


pages/account/tag/[...id].js
```diff
@@ -0,0 +1,267 @@
...
+              X-ADDRESS DETAILS
+            </div>
+
+            <div className="account-tag-xaddress">
+              <b>{originalInput}</b>
+            </div>
+
+            <div className="account-tag-service-description">
+                This address belongs to{' '}
+                <a href={`https://${accountData?.service?.domain || 'crypto.com'}`} target="_blank" rel="noopener noreferrer" className="account-tag-service-link">
```
Comment (ihomp) [inline pages/account/tag/[...id].js:227]: crypto.com?? default? no!

pandablue0809:
@ihomp 

components/Layout/SearchBlock.js
```diff
@@ -153,9 +154,9 @@ export default function SearchBlock({ searchPlaceholderText, tab = null, userDat
   const searchOnChange = (option) => {
     if (!option) return
     if (option.username && !option.username.includes('-')) {
-      onSearch(option.username)
+      onSearch(option.username, option?.tag)
```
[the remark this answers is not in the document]
ihomp: thats why option.username doesn't have the "?"

pages/account/tag/[...id].js
```diff
@@ -0,0 +1,267 @@
...
+    username: accountData?.username,
+    service: accountData?.service?.name,
+    address: accountData?.address || address
+  }
+
+  return (
+    <>
+      <SEO
+        page="Account with Tag"
+        title={`${t('explorer.header.account')} ${userData.service || userData.username || userData.address} - Tag: ${destinationTag}`}
```
Comment (ihomp) [inline pages/account/tag/[...id].js]: "X-Address information"

ihomp (changes requested):
- PayString is not resolved 
/account/username$paystring.crypto.com

- x-Address opens an account page without the tag
/account/XVVFXHFdehYhofb7XRWeJYV6kjTEwbq2mLScCiYyDTHKu9E 

- Direct searching for paystring / xAddress in the landing search - doesn't end up in the right page. (it works if chose from the dropdown)

- we don't need /account/tag route; we need to reuse the /account/ 
- the /account/tag should become a component like Xaddress 

- css need an update, it added some new styles - that are not common in our design. 
- also the social icons shouldn't be grey - it looks like they are disabled/not active

pandablue0809:
@ihomp 
i fixed it.
and i use searchItem because searchFor is always r-address.
so i can't check x-address or paystring using this.
plz review again. 🙏

pages/account/[id].js
```diff
@@ -5,8 +5,9 @@ import axios from 'axios'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import Image from 'next/image'
 import { axiosServer, getFiatRateServer, passHeaders } from '../../utils/axios'
+import { isValidXAddress } from 'ripple-address-codec'
```
Comment (ihomp) [inline pages/account/[id].js]: it's better to import function to check if X-address Valid from our file '../../utils' 

in case we will update it later, so, it will be updated in the whole app at once.

pages/_app.js
```diff
@@ -27,6 +27,7 @@ const WalletConnectModalSign = dynamic(
 import '../styles/globals.css'
 import '../styles/ui.scss'
 import '../styles/components/nprogress.css'
+import '../styles/pages/xaddress-details.scss'
```
ihomp: we should not load css specific for xaddress page to the whole app. 
that css should be loaded on the component level.

ihomp (changes requested):
css file in a wrong dir and should be loaded on the component lvl

pandablue0809:
@ihomp 
i fixed it.
plz review again. 🙏

